PR #4, opened 2022-04-18
fix: prevent account and asset chain mismatch

skeremidchiev:
[LIQ-1238](https://linear.app/liquality/issue/LIQ-1238/boost-wrong-accountid)


monokh:
This change was already made

PR #6, opened 2022-04-20
[Feat]: Migrate NFT Actions, Mutations and State

Jennievon:


src/store/actions/getNFTAssets.js
```diff
@@ -0,0 +1,38 @@
...
+        ? true
+        : false
+    } else {
+      asset['starred'] = false
+    }
+  })
+
+  commit('SET_NFT_ASSETS_NUMBER', nft.assets.length)
+
+  let result = nft.assets.reduce(function (r, a) {
```
monokh: Name `r` and `a` so it's clearer what's happening

src/store/types.ts
```diff
@@ -208,4 +208,126 @@ export interface RootState {
   experiments: Partial<Record<ExperimentType, boolean>>;
   watsNewModalVersion: string;
   enabledChains: WalletIdNetworkMap<ChainId[]>;
+
+  nftAssets: NFTAssets,
+  nftAssetsNumber: number,
+  starredNFTs: NFTAsset[]
+}
+
+export type NFTAsset = {
```
monokh: Why is this type defined in wallet-core? Doesn't it come from CAL/chainify? It should be imported from there

src/store/actions/getNFTAssets.js
```diff
@@ -0,0 +1,38 @@
...
+          item.asset_contract.address === asset.asset_contract.address && item.id === asset.id
+      ).length
+        ? true
+        : false
+    } else {
+      asset['starred'] = false
+    }
+  })
+
+  commit('SET_NFT_ASSETS_NUMBER', nft.assets.length)
```
monokh: There's no need to commit something to the state that can be derived from the state directly.
When you need the length you can do state.nfts.length

src/store/actions/getNFTAssets.js
```diff
@@ -0,0 +1,38 @@
...
+
+  let result = nft.assets.reduce(function (r, a) {
+    r[a.collection.name] = r[a.collection.name] || []
```
monokh: Sorting and grouping should not be in the action and is more of a UI concern.
Rather have all NFTs flat in the structure and sorting/ grouping happening elsewhere. You can add it as a getter if you want - there are similar ones in the getters you can look at

src/store/state.ts
```diff
@@ -49,6 +49,9 @@ const initialState: RootState = {
   experiments: {},
   watsNewModalVersion: '',
   enabledChains: {},
+  nftAssets: {},
+  nftAssetsNumber: 0,
+  starredNFTs: []
```
monokh: There seems to be a lot of complexity introduced because of this array which is separately storing the starred NFTs. It would be better if `starred` was just a property of the NFT, to get a list of starred NFTs, you have to just filter that and that can be done on the UI

src/store/mutations.ts
```diff
@@ -423,6 +425,34 @@ export default {
       ...payload,
     };
   },
+  SET_NFT_ASSETS(state: RootState, payload: NFTAssets) {
+    state.nftAssets = payload
```
monokh: Aren't the nfts different depending on the network and walletid at the very least? what about depending on account? If so, then they need to be split in the state under those categories. See for example the `history` state object

monokh (changes requested):
Please also add a test for the actions that does some basic checks that nfts are retrieved

src/factory/client.ts
```diff
@@ -123,7 +124,7 @@ function createEthereumClient(
   } else {
     ethClient.addProvider(
       new EthereumJsWalletProvider({
-        network: ethereumNetwork,
+        network: { ...ethereumNetwork, chainId: 4 },
```
monokh: Are we really switching to rinkeby? As mentioned this will break thorchain

src/factory/client.ts
```diff
@@ -147,6 +148,9 @@ function createEthereumClient(
   // TODO: remove when Chainify is added
   (ethClient.wallet as any).signTypedMessage = signTypedMessage.bind(ethClient.wallet);
 
+  const openSeaAPI = 'https://rinkeby-api.opensea.io/api/v1/';
```
monokh: Only rinkeby? What about mainnet?

src/factory/client.ts
monokh: maybe you can share the rinkeby version with QA just for testing, but keep ropsten in the official build

src/store/actions/performNextAction/sendNFT.ts
```diff
@@ -0,0 +1,56 @@
...
+    network,
+    walletId,
+    asset: transaction.from,
+    accountId: transaction.accountId,
+  });
+  try {
+    const tx = await client.chain.getTransactionByHash(transaction.txHash);
+    console.log('🚀 ~ file: sendNFT.ts ~ line 30 ~ tx', tx);
+    if (tx && tx.confirmations && tx.confirmations > 0) {
+      dispatch.updateBalances({
```
monokh: Why do you need to update balances? Isn't that just for assets not NFTs?

monokh: Should you update the NFTs of the account?

src/store/actions/sendNFTTransaction.ts
```diff
@@ -0,0 +1,92 @@
...
+    fee,
+    feeLabel,
+    fiatRate,
+    nft,
+  }: // data,
+  {
+    network: Network;
+    accountId: AccountId;
+    walletId: WalletId;
+    contract: Address | string;
```
monokh: Shouldn't these values "contract", "Receiver" "tokenIDs" all come from the `nft` ? Having duplicates of the data can cause problems

src/store/getters.ts
```diff
@@ -282,4 +282,17 @@ export default {
     }
     return false;
   },
+  nftAssetsByCollection(...context: GetterContext): NFTAsset[] {
+    const { state } = rootGetterContext(context);
+    const { nftAssets } = state;
+    const result = nftAssets.reduce(function (assets: any, asset: any) {
```
monokh: No `any` please, the type of these objects should be known

src/store/getters.ts
```diff
@@ -282,4 +282,17 @@ export default {
...
+      assets[asset.collection.name] = assets[asset.collection.name] || [];
+      assets[asset.collection.name].push(asset);
+      assets[asset.collection.name].sort(function (assetA: any, assetB: any) {
+        return assetA.starred === assetB.starred ? 0 : assetA.starred ? -1 : 1;
+      });
+      return assets;
+    }, Object.create(null));
```
monokh: `Object.create(null)` vs `{}`?

src/store/types.ts
```diff
@@ -135,6 +136,18 @@ export interface SendHistoryItem extends BaseHistoryItem {
...
 
+export interface NFTSendHistoryItem extends BaseHistoryItem {
+  type: TransactionType.NFT;
+  toAddress: string;
+  amount: string;
+  tx: Transaction;
+  nft: NFTAsset;
+  txHash: string;
+  accountId: AccountId;
+  fiatRate: number;
```
monokh: is this needed?

src/store/types.ts
```diff
@@ -135,6 +136,18 @@ export interface SendHistoryItem extends BaseHistoryItem {
   status: SendStatus;
 }
 
+export interface NFTSendHistoryItem extends BaseHistoryItem {
+  type: TransactionType.NFT;
+  toAddress: string;
+  amount: string;
```
monokh: does NFT have an amount?

src/store/state.ts
```diff
@@ -49,6 +49,8 @@ const initialState: RootState = {
   experiments: {},
   watsNewModalVersion: '',
   enabledChains: {},
+  nftAssets: [],
+  starredNFTs: [],
```
monokh: I left a comment about not requiring a separate state for starred NFTs and just using the property in `nftAssets`. You can again use a getter if you really need it in a list like this

src/store/mutations.ts
```diff
@@ -427,6 +437,23 @@ export default {
       ...payload,
     };
   },
+  SET_NFT_ASSETS(state: RootState, payload: { nftAssets: NFTAsset[]; network: Network; walletId: WalletId }) {
+    // const walletHistory = state.history[payload.network]?.[payload.walletId];
+    state.nftAssets = payload.nftAssets;
```
monokh: Don't think this comment has been addressed yet: 
https://github.com/liquality/wallet-core/pull/6#discussion_r867830110

We have to figure out handling different chain/network/account now. Once the state is created, it will be difficult to migrate users to different state when we need support for it (which would be very soon)

monokh (changes requested):

src/factory/client.ts
Jennievon: How do we handle it when we need to release though?

src/store/types.ts
Jennievon: By 'this', do you mean `fiatRate`?

src/store/actions/sendNFTTransaction.ts
Jennievon: Yeah, just the `contract` and `tokenIDs` though but noted. 👌🏽

src/store/types.ts
monokh: yes

src/factory/client.ts
monokh: I'm not sure we can support testnet NFTs with the current state of the wallet. The liquidity providers, liquality and thorchain are running on ropsten and we can't sacrifice that for NFTs

We have to have a separate conversation for how to enable multiple testnet networks for the wallet

AndonMitev: I heard somewhere rinkeby will be deprecated

src/store/actions/getNFTAssets.ts
```diff
@@ -0,0 +1,42 @@
...
+  }: {
+    network: Network;
+    walletId: WalletId;
+  }
+): Promise<any> => {
+  const { state, commit, getters } = rootActionContext(context);
+  const client = getters.client({
+    network,
+    walletId,
+    asset: 'ETH',
```
AndonMitev: What if we want to get for polygon for example

src/store/actions/getNFTAssets.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (state.starredNFTs) {
+      asset['starred'] = state.starredNFTs.filter(
+        (item) => item.asset_contract.address === asset.asset_contract.address && item.id === asset.id
+      ).length
+        ? true
+        : false;
+      nftAssets.push(asset);
+    } else {
+      asset['starred'] = false;
+      nftAssets.push(asset);
```
AndonMitev: you can move `nftAssets.push` outside the if else because it's executed anytime anyway

src/store/actions/getNFTAssets.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+  });
+  console.log('🚀 ~ file: getNFTAssets.ts ~ line 34 ~ nftAssets ~ nftAssets', nftAssets);
```
AndonMitev: `console.log`

src/store/actions/performNextAction/sendNFT.ts
```diff
@@ -0,0 +1,50 @@
...
+  const { getters } = rootActionContext(context);
+  const client = getters.client({
+    network,
+    walletId,
+    asset: transaction.from,
+    accountId: transaction.accountId,
+  });
+  try {
+    const tx = await client.chain.getTransactionByHash(transaction.txHash);
+    console.log('🚀 ~ file: sendNFT.ts ~ line 30 ~ tx', tx);
```
AndonMitev: log

bradleySuira (approved):
LGTM

src/store/actions/getNFTAssets.ts
```diff
@@ -22,20 +22,50 @@ export const getNFTAssets = async (
...
-  const nftAssets: NFTAsset[] = [];
-  nft.assets.forEach((asset: any) => {
-    if (state.starredNFTs) {
-      asset['starred'] = state.starredNFTs.filter(
-        (item) => item.asset_contract.address === asset.asset_contract.address && item.id === asset.id
-      ).length
-        ? true
-        : false;
-      nftAssets.push(asset);
+  const oldAssets: NFTAsset[] = state.nftAssets;
```
monokh: What is the purpose of this function?  Can you describe input and outputs?
It's really hard to understand and I feel there is a much clearer way to write what this is doing

monokh (changes requested):

src/store/actions/getNFTAssets.ts
```diff
@@ -0,0 +1,82 @@
...
+  });
+
+  const nft = await client.nft.fetch();
+
+  const account =  state.accounts[walletId]![network].find((a) => a.id === accountId);
+
+  const nftAssetsStoredInState: NFTAsset[] = account?.nftAssets || [];
+  const nftAssetsFetched: NFTAsset[] = nft.assets;
+
+  // set new nft assets to a new map
```
monokh: I see that you added comments but we really need to simplify this function overall
It looks like it's jsut merging 2 arrays and default starred = false. That shouldn't require this much code and should be doable with some map+reduce. Can you help @kraikov ?

kraikov: I think we can simply use the fetched array and set the starred ones from the state.

Jennievon: Thanks, @monokh @kraikov. 👍 
This has been addressed.

src/factory/client.ts
kraikov: We should not switch to Rinkeby. Can you please revert this back.

src/store/actions/getNFTAssets.ts
```diff
@@ -0,0 +1,45 @@
...
+  const nft = await client.nft.fetch();
+
+  const account =  state.accounts[walletId]![network].find((a) => a.id === accountId);
+
+  const nftAssetsStoredInState: NFTAsset[] = account?.nftAssets || [];
+  const nftAssetsFetched: NFTAsset[] = nft.assets;
+
+  const nftAssets: NFTAsset[] = [];
+
+  nftAssetsFetched.forEach((nftAsset) => {
```
kraikov: can you use `map` instead of `forEach`?

src/store/actions/sendNFTTransaction.ts
```diff
@@ -0,0 +1,80 @@
...
+    accountId: AccountId;
+    walletId: WalletId;
+    receiver: string;
+    values: number[];
+    fee: number;
+    feeLabel: FeeLabel;
+    nft: NFTAsset;
+  }
+): Promise<any> => {
+  const asset = 'ETH';
```
kraikov: this one is confusing.... sending an NFT should not contain any `from` and `to` assets. Can we set them to `null` ?

src/store/actions/sendNFTTransaction.ts
```diff
@@ -0,0 +1,80 @@
...
+    network,
+    accountId,
+    walletId,
+    receiver,
+    values,
+    fee,
+    feeLabel,
+    nft,
+  }: {
+    network: Network;
```
kraikov: Can we please have a dedicated type for the input param?

src/store/actions/sendNFTTransaction.ts
```diff
@@ -0,0 +1,80 @@
...
+    accountId: AccountId;
+    walletId: WalletId;
+    receiver: string;
+    values: number[];
+    fee: number;
+    feeLabel: FeeLabel;
+    nft: NFTAsset;
+  }
+): Promise<any> => {
```
kraikov: Avoid using `any`

src/store/getters.ts
```diff
@@ -282,4 +292,36 @@ export default {
...
     return false;
   },
+  nftAssetsByCollection(...context: GetterContext): NFTAsset[] {
+    const { getters } = rootGetterContext(context);
+    const { nftAssetsByAccount } = getters;
+    let nftAssetsByCollection: NFTAsset[] = [];
+    // spread all nfts into one array
+    Object.values(nftAssetsByAccount).forEach((nftAssets) => {
+      nftAssetsByCollection = {
+        ...nftAssetsByCollection,
```
kraikov: `nftAssetsByCollection` is always an empty array, right?

src/store/getters.ts
```diff
@@ -282,4 +292,36 @@ export default {
...
+    accountsData.forEach((account) => {
+      if(account.nftAssets) {
+       const result = account.nftAssets.reduce(function (assets: NFTAssets, asset: NFTAsset) {
+            assets[asset.collection.name] = assets[asset.collection.name] || [];
+            assets[asset.collection.name].push(asset);
+            assets[asset.collection.name].sort(function (assetA: NFTAsset, assetB: NFTAsset) {
+              return assetA.starred === assetB.starred ? 0 : assetA.starred ? -1 : 1;
+            });
+            return assets;
+          }, Object.create({}));
```
kraikov: can't you just use `{}`

src/store/getters.ts
```diff
@@ -282,4 +292,36 @@ export default {
...
+  },
+  nftAssetsByAccount(...context: GetterContext): NFTAssets {
+    const { getters } = rootGetterContext(context);
+    const { accountsData } = getters;
+    const nftAssetsByAccount: NFTAssets = {};
+    accountsData.forEach((account) => {
+      if(account.nftAssets) {
+       const result = account.nftAssets.reduce(function (assets: NFTAssets, asset: NFTAsset) {
+            assets[asset.collection.name] = assets[asset.collection.name] || [];
+            assets[asset.collection.name].push(asset);
```
kraikov: the above 2 lines can be combiled with:

`(assets[asset.collection.name] ||= []).push(asset)`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR_assignment

kraikov (changes requested):
`package-lock.json` must not be included, because wallet-core uses `yarn.lock`

src/store/actions/sendNFTTransaction.ts
```diff
@@ -0,0 +1,80 @@
...
+    fee,
+    tx,
+    nft,
+    txHash: tx.hash,
+    startTime: Date.now(),
+    status: SendStatus.WAITING_FOR_CONFIRMATIONS,
+    accountId,
+    feeLabel,
+  };
+  console.log('🚀 ~ file: sendNFTTransaction.ts ~ line 64 ~ transaction', transaction);
```
kraikov: Remove.

kraikov (changes requested):

src/store/getters.ts
Jennievon: For now but when we have NFTs in different accounts, we would still need to spread them into the array.

kraikov: oh.. got it... then it will be better to use `reduce` and use different name for the function and parameter because it's a little bit confusing

src/store/actions/sendNFTTransaction.ts
kraikov: can we pass the asset as function param, because in the future we would have different NFTs (e.g. Solana)

kraikov: can we pass the asset as function param, because in the future we would have different NFTs (e.g. Solana)

src/store/actions/getNFTAssets.ts
```diff
@@ -0,0 +1,45 @@
...
+  const nftAssetsFetched: NFTAsset[] = nft.assets;
+
+  const nftAssets: NFTAsset[] = [];
+
+  nftAssetsFetched.map((nftAsset) => {
+    const nftAssetStoredInState = nftAssetsStoredInState.find((a) => a.id === nftAsset.id);
+    if (nftAssetStoredInState) {
+      nftAsset.starred = nftAssetStoredInState.starred;
+    }
+    nftAssets.push(nftAsset);
```
kraikov: the map function populates a new array with the returned value, so there is no need to do 
`nftAssets.push(nftAsset);`

You should do:

`const nftAssets =  nftAssetsFetched.map((nftAsset) => .... return nftAsset)`

kraikov, line 39: the map function populates a new array with the returned value, so there is no need to do 
`nftAssets.push(nftAsset);`

You should do:

`const nftAssets =  nftAssetsFetched.map((nftAsset) => .... return nftAsset)`

src/store/getters.ts
```diff
@@ -298,4 +308,34 @@ export default {
     }
     return false;
   },
+  nftAssetsByCollection(...context: GetterContext): NFTAsset[] {
+    const { getters } = rootGetterContext(context);
+    const { nftAssetsByAccount } = getters;
+    let nftAssetsByCollection: NFTAsset[] = [];
+    nftAssetsByCollection = Object.values(nftAssetsByAccount).reduce((accum, nftAssets) => {
```
kraikov: The `reduce` function is not used correctly here. Please check https://codeburst.io/how-to-reduce-array-to-a-single-type-11cfe2bcae66

kraikov: The `reduce` function is not used correctly here. Please check https://codeburst.io/how-to-reduce-array-to-a-single-type-11cfe2bcae66

src/store/getters.ts
kraikov: This is not yet resolved

kraikov: This is not yet resolved

src/factory/client/clients.ts
```diff
@@ -87,11 +87,14 @@ export function createEVMClient(
...
   mnemonic: string,
   accountType: AccountType,
-  derivationPath: string
+  derivationPath: string,
+  nftProvider?: EvmNftProvider
 ) {
   // disable multicall for Testnets
   const chainProvider = new EvmChainProvider(ethereumNetwork, undefined, feeProvider, !ethereumNetwork.isTestnet);
   const swapProvider = new EvmSwapProvider({ contractAddress: HTLC_CONTRACT_ADDRESS });
+  const walletOptions = { derivationPath, mnemonic };
```
monokh: Unrelated change?

src/factory/client/evm.ts
```diff
@@ -1,25 +1,33 @@
...
+    responseType: 'text',
+    transformResponse: undefined,
+    headers: {
+      'X-Api-Key': '',
+    },
+  };
   const feeProvider = new EIP1559FeeProvider(ethNetwork.rpcUrl as string);
-  return createEVMClient(ethNetwork, feeProvider, mnemonic, accountType, derivationPath);
+  const chainProvider = new EvmChainProvider(
+    { ...ethNetwork, chainId: 4 },
```
monokh: As mentioned before, please let the wallet stay on ropsten testnet

src/store/actions/getNFTAssets.ts
```diff
@@ -0,0 +1,47 @@
...
+  {
+    walletId,
+  }: {
+    walletId: WalletId;
+  }
+): Promise<void> => {
+  const { state, commit, getters } = rootActionContext(context);
+  const { networks, nftAssets } = buildConfig;
+
+  networks.forEach((network) => {
```
monokh: We shouldn't update for all networks when the user is not on it. It will slow down the wallet for no reason, 99% only use mainnet. the active network should be passed into this function

src/store/actions/getNFTAssets.ts
```diff
@@ -0,0 +1,47 @@
...
+    const assetKeys = nftAssets[network];
```
monokh: assetKeys? Isn't it chains?

src/store/actions/getNFTAssets.ts
```diff
@@ -0,0 +1,47 @@
...
+    assetKeys.forEach(async (chain) => {
+      const asset = chains[chain].nativeAsset;
+      const account = state.accounts[walletId]![network].find((a) => a.assets.includes(asset));
```
monokh: Why do you need to find the account by the asset? You are looking for the account for that chain and accounts have a `chain` property

monokh: There can be multiple accounts for a chain, each account (therefore address) has it's own list of nfts. You need to make sure every account is processed

src/store/actions/getNFTAssets.ts
```diff
@@ -0,0 +1,47 @@
...
+      const _client = getters.client({
+        network,
+        walletId,
+        asset,
+        accountId: account?.id,
+      });
+
+      const nft: any = await _client.nft?.fetch();
```
monokh: why do we type this as any? I believe it does return a type

src/store/actions/getNFTAssets.ts
```diff
@@ -0,0 +1,47 @@
...
+
+      const nftAssetsStoredInState: NFTAsset[] = account?.nftAssets || [];
+      const nftAssetsFetched: NFTAsset[] = nft.assets;
+
+      const nftAssets: NFTAsset[] = nftAssetsFetched.map((nftAsset) => {
+        const nftAssetStoredInState = nftAssetsStoredInState.find((asset) => asset.id === nftAsset.id);
+        if (nftAssetStoredInState) {
+          nftAsset.starred = nftAssetStoredInState.starred || false;
+          nftAsset.asset = nftAssetStoredInState.asset || asset;
```
monokh: These 2 statements seem weird, asset and chain are static aren't they? they should always be the same so why the  || ?
Also, why store asset? an nft is for a chain, doesn't have anything to do with an asset i believe

src/store/migrations/before_after_migration_script.test.ts
```diff
@@ -0,0 +1,116 @@
+import * as Process from 'process';
```
monokh: unrelated?

src/store/types.ts
monokh: ?

monokh (changes requested):

src/store/actions/getNFTAssets.ts
Jennievon: `void` is being set as the type in `chainify` for the `fetch` response, hence, we cannot set a specific type as it isn't assignable to `void`. That's the reason I used `any`

src/factory/client/evm.ts
Jennievon: This wouldn't be merged. It's just to test the NFTs on testnet.

src/store/migrations/before_after_migration_script.test.ts
Jennievon: some file changes are from the `master` branch

monokh: does that mean chainify branch needs to be updated to include something? This PR should not be adding anything unrelated

src/store/actions/getNFTAssets.ts
monokh: Yea it's not good that chainify is returning any 
Is the actual type `NftAsset` that you have in this project? In that case we can move it to chanify so you get the type here

monokh:
Closing to reopen because github actions behaving funny

PR #12, opened 2022-04-27
feat: integrate chainify

kraikov:
TODO: test the wallet core along with the extension


src/factory/clients.ts
```diff
@@ -0,0 +1,140 @@
...
+  mnemonic: string,
+  accountType: AccountType,
+  baseDerivationPath: string
+) {
+  const isTestnet = network === 'testnet';
+  const bitcoinNetwork = ChainNetworks.bitcoin[network];
+  const esploraApi = buildConfig.exploraApis[network];
+  const batchEsploraApi = buildConfig.batchEsploraApis[network];
+
+  const feeProvider = new BitcoinFeeApiProvider('https://liquality.io/swap/mempool/v1/fees/recommended');
```
monokh: Shouldn't the fee provider different for bitcoin testnet?
Those feels i believe are mainnet fees

src/factory/clients.ts
```diff
@@ -0,0 +1,140 @@
...
+    };
+    const walletProvider = new BitcoinHDWalletProvider(walletOptions, chainProvider);
+    swapProvider.setWallet(walletProvider);
+  }
+
+  if (isTestnet) {
+    chainProvider.setFeeProvider(null);
+  }
+
+  return new Client().connect(swapProvider);
```
monokh: Kinda weird how only swap provider can use connect here
in the ideal way, none of the providers need to take other providers as argument, or need to set it eg.`setFeeProvider`

at the end you just do `return new Client().connect(swapProvider).connect(feeProvider).connect(chainProvider).connect(walletProvider) and that's it

src/factory/index.ts
```diff
@@ -0,0 +1,48 @@
+import { ChainId } from '@liquality/cryptoassets';
```
monokh: I think this should be in it's own module (like `client.ts` or `client/index.ts`), not in index. 
This package (`factory`) is also has swapProvider factory.

src/swaps/fastbtc/FastbtcSwapProvider.ts
```diff
@@ -80,6 +83,13 @@ class FastbtcSwapProvider extends SwapProvider {
     });
   }
 
+  public getClient(network: Network, walletId: string, asset: string, accountId: string) {
+    return super.getClient(network, walletId, asset, accountId) as Client<
```
monokh, line 90: hmm, isn't this also called to create an RSK client ? Then why would it be bitcoin type only

src/swaps/liquality/LiqualitySwapProvider.ts
```diff
@@ -506,39 +510,6 @@ export class LiqualitySwapProvider extends SwapProvider {
     };
   }
 
-  private async fundSwap({ swap, network, walletId }: NextSwapActionRequest<LiqualitySwapHistoryItem>) {
```
monokh, line 505: As mentioned offline, we need the approval step separately in order to track it.

src/factory/clients.ts
kraikov: `new Client().connect(swapProvider);` is the same as 
`new Client().connect(swapProvider).connect(chainProvider).connect(walletProvider)`

src/factory/clients.ts
kraikov: this is handled implicitly :(
https://github.com/liquality/wallet-core/commit/6d7bc105d24820cf04c6e6c8cf62fbc1d4889d1e#diff-48b303d10d6d68390a3415e1901cf69f3b382c2e0005e3b28b86ad90ea404e34R75

src/factory/clients.ts
kraikov: in that case there is no need to use

`new Client().connect(swapProvider).connect(chainProvider).connect(walletProvider)` because both chain and wallet providers are already available

src/swaps/fastbtc/FastbtcSwapProvider.ts
kraikov: was not aware... will fix that.. however I only did it because of "getTotalFees" which is BTC specific

edit: as far as I can tell getClient is only used for BTC...can you confirm/disprove ?

src/factory/clients.ts
kraikov: maybe reverse the logic?

i.e. 
`if mainnet => setFeeProvider`

monokh: yea I think that makes more sense that way

monokh (approved):

bradleySuira (approved):
LGTM

package.json
```diff
@@ -119,5 +91,21 @@
     "prepack": "pinst --disable",
     "postpack": "pinst --enable"
   },
-  "packageManager": "yarn@3.1.1"
+  "packageManager": "yarn@3.1.1",
+  "resolutions": {
```
AndonMitev: Do we need resolutions, is this for the new linking? Asking because I can see it's referring to you local space or it's just a example how it must look like

src/swaps/astroport/AstroportSwapProvider.ts
```diff
@@ -188,12 +189,8 @@ class AstroportSwapProvider extends SwapProvider {
...
 
   _getRPC() {
-    const { chainID, nodeUrl } = TerraNetworks.terra_mainnet;
-
-    return new LCDClient({
-      chainID,
-      URL: nodeUrl,
-    });
+    const { chainId, rpcUrl } = TerraNetworks.terra_mainnet;
+    return new LCDClient({ chainID: String(chainId), URL: String(rpcUrl) });
```
AndonMitev, line 193: I think we can remove `String(...)` from `chainId` and `rpcUrl` because they are strings by default: https://github.com/liquality/chainify/blob/7fc75bbd9a603c7fd38181a07a4e2e8ce5a67906/packages/terra/lib/networks.ts#L3

src/swaps/astroport/AstroportSwapProvider.ts
```diff
@@ -136,11 +136,12 @@ class AstroportSwapProvider extends SwapProvider {
...
       const tx = await client.chain.getTransactionByHash(swap.swapTxHash);
       if (tx && tx.confirmations && tx.confirmations > 0) {
-        const { status } = await client.getMethod('getTransactionByHash')(swap.swapTxHash);
+        const { status } = tx;
         this.updateBalances(network, walletId, [swap.from]);
+
         return {
           endTime: Date.now(),
-          status,
+          status: status === TxStatus.Success ? 'SUCCESS' : 'FAILED',
```
AndonMitev, line 144: I think we can remove this check here and just leave it `status`, if status is equal to TxStatus.Success for example that means `status` will equal to `SUCCESS`:

https://github.com/liquality/chainify/blob/8257ba10caa513086c484219027a49756d6e3228/packages/types/lib/Transaction.ts#L40

src/swaps/astroport/AstroportSwapProvider.ts
```diff
@@ -136,11 +136,12 @@ class AstroportSwapProvider extends SwapProvider {
     try {
       const tx = await client.chain.getTransactionByHash(swap.swapTxHash);
       if (tx && tx.confirmations && tx.confirmations > 0) {
-        const { status } = await client.getMethod('getTransactionByHash')(swap.swapTxHash);
+        const { status } = tx;
```
AndonMitev, line 139: TOP 😂

src/swaps/liquality/LiqualitySwapProvider.ts
```diff
@@ -127,13 +128,22 @@ export class LiqualitySwapProvider extends SwapProvider {
     };
   }
 
-  public async newSwap({ network, walletId, quote: _quote }: SwapRequest<LiqualitySwapHistoryItem>) {
+  public async newSwap(swapRequest: SwapRequest<LiqualitySwapHistoryItem>) {
+    const approveTx = await this.approve(swapRequest, true);
+    const updates = approveTx !== null ? approveTx : await this.initiateSwap(swapRequest);
```
AndonMitev: I think this can be written: `const updates = approveTx ? approveTx : await this.initiateSwap(swapRequest);`

src/swaps/astroport/AstroportSwapProvider.ts
kraikov: image
image

src/swaps/astroport/AstroportSwapProvider.ts
kraikov: The issue is that there is no concept of `PENDING` that's why we either need success or failed

src/swaps/liquality/LiqualitySwapProvider.ts
kraikov: or even better

` const updates = approveTx || (await this.initiateSwap(swapRequest));`

PR #15, opened 2022-05-04
feat: Integrate Anchor UST Staking

AndonMitev:


src/earn/anchor/AnchorEarnProvider.ts
```diff
@@ -0,0 +1,54 @@
...
+class AnchorEarnProvider extends EarnProvider {
+  client: AnchorEarn;
+
+  constructor({ network }: { network: string }) {
+    super();
+
+    this.client = new AnchorEarn({
+      chain: CHAINS.TERRA,
+      network: network === 'mainnet' ? NETWORKS.COLUMBUS_5 : NETWORKS.BOMBAY_12,
+      mnemonic: store.state.wallets[0].mnemonic,
```
monokh, line 16: hmm this might be a problem, what about the account?

src/store/actions/getApy.ts
```diff
@@ -0,0 +1,13 @@
+import { ActionContext } from '..';
```
monokh: Put these actions into separate folder? OR export all of them from same file? `earn.ts`

monokh (changes requested):

src/store/actions/getApy.ts
AndonMitev: Resolved

src/earn/anchor/AnchorEarnProvider.ts
AndonMitev: What do you mean by "account"? Aren't they based on the same mnemonic?

monokh: They are based on the mnemonic, but the question is, if a user has 3 terra accounts, which is used? The account that is used for the anchor earning should be user specified

AndonMitev: Hmm trying to wrap my head around 😆 . Do we really care about which account is used because the AnchorClient is created by the mnemonic and not public address OR you count as a account `store.state.wallets[1].mnemonic` and then `store.state.wallets[2].mnemonic`  OR i do get this completely wrong :( .

monokh: A user's wallet has multiple accounts. On each account, the funds are separate right? When you want to stake, which account will be used? Here we are only specifying the mnemonic. What if we have 3 accounts, each with 10 UST. Which is used?

AndonMitev: gotcha, no idea right now later when we integrate it into the wallet i guess we can understand and adjust it cuz really not sure rn

monokh:
Lets bring this back when we can do earn again. Or if you can take out the anchor stuff
Main worry is the anchor sdk dependency

PR #18, opened 2022-05-10
feat: add support for signing typed data (move to chainify later)

kraikov:


monokh (approved):

src/utils/eth_signTypedData_v4.ts
```diff
@@ -0,0 +1,52 @@
...
+  const method = payload.method;
+  const first = payload.params[0];
+  const second = payload.params[1];
+
+  const msgParams = { privateKey, version: methodToVersion[method] } as SignTypedMessageType<
+    SignTypedDataVersion,
+    MessageTypes
+  >;
+
+  if (isAddress(first)) {
```
monokh, line 32: This feels a bit dangerous, Is this issue documented somewhere?
Is it based on the version of `eth_signTypedData` ? If so, it's best to explicitly handle each type rather than allowing anything through

AndonMitev:
LGTM 🚀

PR #19, opened 2022-05-10
Liq 1100 integrate hop exchange

tima-t:
📖  liq-1100 📖


src/store/types.ts
```diff
@@ -147,6 +148,9 @@ export interface SwapHistoryItem extends BaseHistoryItem {
   toAmount: string;
   bridgeAsset?: Asset;
   path?: string[];
+  hopAsset?: any;
```
monokh: These should be in the `HopSwapHistoryItem` not the base
see thorchain for example

src/swaps/hop/HopSwapProvider.ts
```diff
@@ -0,0 +1,579 @@
...
+   */
+  // eslint-disable-next-line no-unused-vars
+  async getSupportedPairs() {
+    return [];
+  }
+
+  confirmationsSendCount(networkName: string) {
+      const networkToConfirmations: { [key: string]: number } =    
+      {
+        ethereum: 1,
```
monokh: How are these determined? If it's a hop parameter, they should provide this via a library?
If it's liquality defined then this is a chain configuration that can be retrieved from `cryptoassets`. It's `chains[chain].safeConfirmations`

src/swaps/hop/HopSwapProvider.ts
```diff
@@ -0,0 +1,579 @@
...
+        arbitrum: 20,
+        polygon: 256
+      }
+      return networkToConfirmations[networkName]
+  }
+
+  gasLimit(networkName: string){
+    const networkToGasLimit : { [key: string]: any } = {
+      arbitrum: {
```
monokh, line 72: Why hardcoded? we cannot estimate these using rpc?

src/swaps/hop/HopSwapProvider.ts
```diff
@@ -0,0 +1,579 @@
...
+      {
+        ethereum: 1,
+        arbitrum: 20,
+        polygon: 256
+      }
+      return networkToConfirmations[networkName]
+  }
+
+  gasLimit(networkName: string){
+    const networkToGasLimit : { [key: string]: any } = {
```
monokh: return actual type, not any

src/swaps/hop/HopSwapProvider.ts
```diff
@@ -0,0 +1,579 @@
...
+      ethereum: {
+        send: 150000,
+        approve: 100000
+      }
+    }
+    return networkToGasLimit[networkName]
+  }
+
+  getChain(chainName: string){
+    const slugToChain : { [key: string]: any } = {
```
monokh: return actual type, not any

src/swaps/hop/HopSwapProvider.ts
```diff
@@ -0,0 +1,579 @@
...
+  }
+
+  _getDestinationTxGQL(transferId: string, recipient: string, isFromL1Source: boolean) {
+    return isFromL1Source
+      ? this.GQL_getDestinationTxHashFromL1Source(recipient)
+      : this.GQL_getDestinationTxHashFromL2Source(transferId)
+  }
+
+  _getHop(network: Network, signer = undefined) {
+    if (!network) return null
```
monokh: how can there be no network if network parameter is mandatory?

src/swaps/hop/HopSwapProvider.ts
```diff
@@ -0,0 +1,579 @@
...
+    return new Wallet(privKey, provider)
+  }
+
+  async _getBridgeWithSigner(hopAsset: any, hopChainFrom: any, network: Network, walletId: string, from: string, fromAccountId: string) {
+    const chainFrom = this.getChain(hopChainFrom.slug)
+    const client = this._getClient(network, walletId, from, fromAccountId)
+    const privKey = await client.wallet.exportPrivateKey()
+    const hop = this._getHop(network)
+    const signer = new Wallet(privKey, hop?.getChainProvider(chainFrom))
+    const bridge = hop?.connect(signer).bridge(hopAsset)
```
monokh: All these question marks are making me worried. In this case, hop could be undefined and bridge could be undefined, but the code is triggered and will continue. Probably somewhere way before this code gets triggered, you can make some basic checks and exist early

src/swaps/hop/HopSwapProvider.ts
```diff
@@ -0,0 +1,579 @@
...
+    return bridge
+  }
+
+  _findAsset(asset: any, chain: any, tokens: any, tokenName: string) {
```
monokh: no any pls

src/swaps/hop/HopSwapProvider.ts
```diff
@@ -0,0 +1,579 @@
...
+  // eslint-disable-next-line no-unused-vars
+  public async getQuote({ network, from, to, amount } : QuoteRequest) {
+    if (amount <= new BN(0)) return null
+    const assetFrom = cryptoassets[from]
+    const assetTo = cryptoassets[to]
+    const fromAmountInUnit = currencyToUnit(cryptoassets[from], new BN(amount))
+    const hop = this._getHop(network)
+    if (!hop || !hop.isValidChain(assetFrom.chain) || !hop.isValidChain(assetTo.chain)) return null
+    const info = this._getSendInfo(assetFrom, assetTo, hop)
+    if (!info?.bridgeAsset || !info?.chainFrom || !info?.chainTo) return null
```
monokh: Isn't all these already checked in getSendInfo and returning null? Why are we checking each property again? And they shouldn't be optional then if they are required for the swap.

src/swaps/hop/HopSwapProvider.ts
```diff
@@ -0,0 +1,579 @@
...
+    const { bridgeAsset, chainFrom, chainTo } = info
+    const bridge = hop.bridge(bridgeAsset)
+    const sendData = await bridge.getSendData(fromAmountInUnit.toString(), chainFrom, chainTo)
+    if (!sendData) return null
+    return {
+      from,
+      to,
+      // Amounts should be in BigNumber to prevent loss of precision
```
monokh: remove this comment, we use string now to avoid loss in precision

src/swaps/hop/HopSwapProvider.ts
```diff
@@ -0,0 +1,579 @@
...
+        .plus(new BN(sendData.adjustedDestinationTxFee.toString()))
+        .toString()
+    }
+  }
+
+  async _approveToken(bridge: any, chainFrom: any, fromAmount: any, signer: any, fee: any) {
+    const txData = await bridge.populateSendApprovalTx(fromAmount, chainFrom)
+    const approveTx = await signer.sendTransaction({
+      ...txData,
+      gasPrice:
```
monokh: only gas price? What if it's EIP1559? Not sure how other swap providers work in this regard

src/swaps/hop/HopSwapProvider.ts
```diff
@@ -0,0 +1,579 @@
...
+          amount
+          transactionHash
+          token
+          timestamp
+        }
+      }
+    `
+  }
+  // L1->L2
+  GQL_getDestinationTxHashFromL1Source(recipient : string) {
```
monokh: This type of naming seems weird, perhaps have another utility for these, Like what the astroport provider does

monokh (changes requested):

monokh:
Ser pls typescript all the things

src/store/types.ts
tima-t: I agree, I will update it

src/swaps/hop/HopSwapProvider.ts
tima-t: chains[chain].safeConfirmations will do the job. I will remove the other function

src/swaps/hop/HopSwapProvider.ts
tima-t: It is done like this because 
- when you want to call estimate fee on the send method it throws an error because it cannot be estimated, because the approve is still not done (maybe the problem was appearing on Arbitrum afaik)
- their team suggested these values, so I used them

The problem is similar to  this one, tho 
https://github.com/liquality/wallet-core/blob/master/src/swaps/sovryn/SovrynSwapProvider.ts#L296

src/swaps/hop/HopSwapProvider.ts
tima-t: All the other amms like uniswap, oneinch, sovryn are working the same way - only gas price

src/swaps/hop/HopSwapProvider.ts
tima-t: I moved the queries to separate file

src/swaps/hop/queries.ts
```diff
@@ -0,0 +1,62 @@
+// L1->L2
+function  queryGetDestinationTxHashFromL1Source(recipient : string) {
```
monokh: shouildn't need query prefix?

package.json
```diff
@@ -19,6 +19,7 @@
   "dependencies": {
     "@blobfishkate/sovryncontracts": "1.3.27",
     "@ethersproject/address": "^5.6.0",
+    "@hop-protocol/sdk": "^0.0.1-beta.314",
```
monokh: please pin the version

monokh (approved):

PR #24, opened 2022-05-13
Feature ledger providers

bradleySuira:
This PR includes:
- a fix for what's new modal 
- getLedgerAccounts updates to work with the latest in the wallet
- add a migration to remove existing ledger accounts to clean the sate and start with the new accounts created
- add chainCode and publicKey to the client getter and factory to use it in bitcoin ledger provider
- remove all related to ledger bridge that we use


src/factory/client.ts
```diff
@@ -436,21 +438,34 @@ export const createClient = (
   network: Network,
   mnemonic: string,
   accountType: AccountType,
-  derivationPath: string
+  derivationPath: string,
+  publicKey?: string,
+  chainCode?: string
 ) => {
   const assetData = cryptoassets[asset];
```
monokh, line 444: Can you do a check here that throws an error when publickey, chaincode is provided but client is not BTC?

src/factory/client.ts
```diff
@@ -436,21 +438,34 @@ export const createClient = (
   network: Network,
   mnemonic: string,
   accountType: AccountType,
-  derivationPath: string
+  derivationPath: string,
+  publicKey?: string,
```
monokh: better put these 2 params into one object:

```
derivationPath: string,
bitcoinLedgerCache?: { publicKey: string, chainCode: string }
```

This makes the purpose of these options much clearer

src/store/state.ts
```diff
@@ -47,7 +47,7 @@ const initialState: RootState = {
     notAskAgain: false,
   },
   experiments: {},
-  watsNewModalVersion: '',
+  whatsNewModalVersion: '',
```
monokh, line 50: You have changed this but there is no migration, This means there will be an old state in the wallet called `watsNewModalVersion` 
Will this cause any bugs? I am happy without migration if it does not introduce errors when wallet updates.

src/store/getters.ts
```diff
@@ -72,6 +74,8 @@ export default {
       useCache?: boolean;
       accountType?: AccountType;
       accountIndex?: number;
+      chainCode?: string | undefined;
```
monokh: `chainCode?: string | undefined`
is the same as `chainCode?: string` ? 
`?` includes undefined

Also make sure this parameter matches the same format as in `client.ts`

src/store/migrations/19_remove_existing_ledger_accounts.ts
```diff
@@ -0,0 +1,25 @@
+import { Account, AccountType } from "../types";
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export const removeExistingLedgerAccounts = {
```
monokh, line 4: Can you add a test for the migration :)
You can copy a wallet state from pre this version and test it

monokh (changes requested):

src/factory/client.ts
bradleySuira: done

src/store/getters.ts
bradleySuira: fixed

src/store/state.ts
bradleySuira: added inside the last migration now

src/factory/client.ts
bradleySuira: it works, tested with the wallet and evm accounts

src/store/migrations/19_remove_existing_ledger_accounts.ts
bradleySuira: @monokh added this way to test the state, also you can run it `npx ts-node ./src/store/migrations/19_remove_existing_ledger_accounts_test.ts` but to be honest I think that can be better if I try later to find a way to test all migrations and don't need to upload any json files or maybe we can set a folder for all states and examples per version.

Screen Shot 2022-05-25 at 7 27 45 AM

src/factory/client.ts
monokh: I'm sure it works, but my suggetion gives some feedback to a consumer that would call those parameters on a non BTC client, and expect it to work.

src/store/state.ts
monokh: I dont see this variable being treated in the migration 😄  did you commit?

monokh (approved):

PR #40, opened 2022-05-19
fix: bugs related to fees on eth - eip1559

tima-t:
📖  LIQ-1368/ 📖

### DIFF
--- a/src/swaps/hop/HopSwapProvider.ts
+++ b/src/swaps/hop/HopSwapProvider.ts
@@ -68,7 +68,7 @@ class HopSwapProvider extends SwapProvider {
   }
 
   gasLimit(networkName: string){
-    const networkToGasLimit : { [key: string]: {send: number, approve: number} } = {
+    const networkToGasLimit : { [key: string]: {[key: string]: number} } = {
       arbitrum: {
         send: 900000,
         approve: 1000000
@@ -78,8 +78,8 @@ class HopSwapProvider extends SwapProvider {
         approve: 300000
       },
       ethereum: {
-        send: 150000,
-        approve: 100000
+        send: 300000,
+        approve: 300000
       }
     }
     return networkToGasLimit[networkName]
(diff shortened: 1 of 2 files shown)


src/swaps/hop/HopSwapProvider.ts
```diff
@@ -195,11 +195,26 @@ class HopSwapProvider extends SwapProvider {
     }
   }
 
+  _formatFee(fee: any, networkName: string, type: string ){
```
monokh: type pls. It should be something like `number | EIP1559Fee` 
`EIP1559Fee` can probably be imported from chainify

src/swaps/hop/HopSwapProvider.ts
```diff
@@ -195,11 +195,26 @@ class HopSwapProvider extends SwapProvider {
...
+    if(fee.maxFeePerGas && fee.maxPriorityFeePerGas){
```
monokh: If you have type, you can discriminate the type ( i think)

monokh (changes requested):

src/swaps/hop/HopSwapProvider.ts
tima-t: I've tried it EIP1559Fee, but it is throwing an error saying something like 

"number cannot have property maxFeePerGas"

So looks like these arbitrary types number I EIP1559Fee causes some problems on compilation. 

Maybe I am missing something but it behaves like this on my side

src/swaps/hop/HopSwapProvider.ts
tima-t: This type I am using for  "approve" or "send" it is different. 

I am not sure if the comment is related to that

src/swaps/hop/HopSwapProvider.ts
tima-t: Maybe in the next version we should think of a way to handle eip1559Fee and Regular gasPrice in some more clever way, because passing arbitrary types is causing warning on TS side

monokh: Let me DM you see if we can figure it out

src/swaps/hop/HopSwapProvider.ts
tima-t: Fixed, thanks for the help!

monokh (approved):

PR #45, opened 2022-05-25
fix: added fee multiplier for 1inch on arbitrum

tima-t:
📖  LIQ-1260 📖

### DIFF
--- a/src/swaps/oneinch/OneinchSwapProvider.ts
+++ b/src/swaps/oneinch/OneinchSwapProvider.ts
@@ -39,7 +39,7 @@ const chainToGasMultiplier: { [chainId: number]: number } = {
   56: 1.5,
   137: 1.5,
   43114: 1.5,
-  42161: 6
+  42161: 10 
 };
 
 export interface OneinchSwapHistoryItem extends SwapHistoryItem {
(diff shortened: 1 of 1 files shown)


AndonMitev (approved):
LGTM 🚀

src/swaps/oneinch/OneinchSwapProvider.ts
```diff
@@ -31,7 +31,15 @@ const chainToRpcProviders: { [chainId: number]: string } = {
...
-  42161: `https://arbitrum-mainnet.infura.io/v3/${buildConfig.infuraApiKey}`,
+  42161: `https://arbitrum-mainnet.infura.io/v3/${buildConfig.infuraApiKey}`
+};
+
+const chainToGasMultiplier: { [chainId: number]: number } = {
+  1: 1.5,
+  56: 1.5,
+  137: 1.5,
+  43114: 1.5,
+  42161: 10
```
AndonMitev, line 42: Looking good, just wondering maybe we can leave only `42161` here as a key cuz other have the same values 1.5 and below `.times(chainToGasMultiplier[chainId] || 1.5)` default value is 1.5 again so really if the key is `42161` only then it might has some difference but it's looking good not something important

monokh (approved):

src/swaps/oneinch/OneinchSwapProvider.ts
tima-t: Yes, I agree. My ideas was that we might need to change this multiplier for some other chain in the future.

bradleySuira (approved):
LGTM

PR #53, opened 2022-06-01
test: added migration tests

beemi:
- Migrations tests to validate mainet & testnet accounts


src/store/migrations/before_after_migration_script.test.ts
```diff
@@ -0,0 +1,125 @@
+import * as Process from 'process';
+import {setupWallet} from '../../index';
+import defaultWalletOptions from '../../walletOptions/defaultOptions';
+import {LATEST_VERSION} from "./index";
+
+test('Upgrade migrations from wallet versions-2 test, should be able validate accounts-mainnet', async () => {
```
kraikov: We need to call [processMigrations](https://github.com/liquality/wallet-core/blob/master/src/store/migrations/index.ts#L55) in order to simulate wallet update. So instead of calling `setupWallet` twice, we have to call it only once and after that we need `processMigrations`

beemi: Done, changed pushed using `isMigrationNeeded` function

kraikov:
LGTM.

kraikov (approved):

PR #55, opened 2022-06-07
feat: ledger

bradleySuira:
- Added changes to client factory to seed ledger accounts


src/factory/client/clients.ts
```diff
@@ -136,17 +133,25 @@ export function createEVMClient(
...
+        address: accountInfo.address
+      });
+    }
     walletProvider = new EvmLedgerProvider(
       {
         network: ethereumNetwork,
-        derivationPath,
+        derivationPath: accountInfo.derivationPath,
+        addressCache,
         transportCreator: ledgerTransportCreator,
```
monokh: We can't asume that all clients will be able to support WebHID transport. 
This should be an option in `WalletOptions` similar to how ledger providers were handled before

src/factory/client/clients.ts
```diff
@@ -91,19 +89,19 @@ export function createBtcClient(
...
       {
         network: bitcoinNetwork,
         addressType,
-        baseDerivationPath,
-        basePublicKey: account?.publicKey,
-        baseChainCode: account?.chainCode,
+        baseDerivationPath: accountInfo.derivationPath,
+        basePublicKey: accountInfo?.publicKey,
+        baseChainCode: accountInfo?.chainCode,
         transportCreator: ledgerTransportCreator,
```
monokh: Same story as other comment. Can't assume this transport will work on mobile

monokh (changes requested):

src/factory/client/clients.ts
bradleySuira: in this case seems like wallet core can handle using it and it's implemented inside the '@chainify/hw-ledger' in `WebHidTransportCreator` I will add it then to the wallet and send inside `WalletOptions` like you said

bradleySuira: fixed

src/factory/client/clients.ts
bradleySuira: fixed

monokh (approved):

PR #60, opened 2022-06-10
Feat/mob 9

adilmezghouti:
- Changed the type of the returned type of the sendTransaction function to be of type SendHistory to make the processing of the send and swap transaction easier and dry
- Casted property to ChainId to remove a warning. We need to make chainify and wallet-core play nice with each other

### DIFF
--- a/src/store/actions/sendTransaction.ts
+++ b/src/store/actions/sendTransaction.ts
@@ -74,5 +74,5 @@ export const sendTransaction = async (
 
   createHistoryNotification(transaction);
 
-  return tx;
+  return transaction;
 };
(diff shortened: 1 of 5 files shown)


monokh (approved):

src/store/actions/sendTransaction.ts
```diff
@@ -40,7 +41,7 @@ export const sendTransaction = async (
 
   const _asset = assetsAdapter(asset)[0];
   const tx = await client.wallet.sendTransaction({
-    to: chains[_asset.chain].formatAddress(to),
+    to: chains[_asset.chain as ChainId].formatAddress(to),
```
monokh, line 43: weird, it actually is `ChainId` type already

kraikov:
Yes, we need to refactor `cryptoassets` so we can get rid of the `assetsAdapter`

PR #62, opened 2022-06-13
feat: add optimism

skeremidchiev:
[LIQ-1310](https://linear.app/liquality/issue/LIQ-1310/integrate-optimism-as-evm-chain-in-chainify)


src/store/migrations/21_enable_optimism_chain.ts
```diff
@@ -0,0 +1,69 @@
+import { ChainId, chains } from '@liquality/cryptoassets';
+import { accountCreator, getNextAccountColor } from '../../utils/accounts';
+import { getDerivationPath } from '../../utils/derivationPath';
+import { Networks } from '../../utils/networks';
+import { AccountType } from '../types';
+
+export const enableOptimismChain = {
+  version: 21,
+  migrate: async (state: any) => {
```
monokh, line 9: Any chance that instead of copy pasting this we can create a utility in migrations that adds an account, then reuse it?

src/utils/fees.ts
```diff
@@ -17,11 +18,24 @@ const feePriceInUnit = (asset: Asset, feePrice: number) => {
...
-function getSendFee(asset: Asset, feePrice: number) {
+function getSendFee(asset: Asset, feePrice: number, l1FeePrice?: number, network?: Network) {
   const assetInfo = cryptoassets[asset];
-  const fee = new BN(assetInfo.sendGasLimit).times(feePriceInUnit(asset, feePrice));
-
-  return unitToCurrency(cryptoassets[getNativeAsset(asset)], fee);
+  const nativeAsset = cryptoassets[getNativeAsset(asset)];
+
+  if (l1FeePrice && isMultiLayeredChain(assetInfo.chain)) {
+    if (assetInfo.chain === ChainId.Optimism) {
```
monokh, line 26: Do you need to check both `isMultiLayeredChain` and the chain? Since you know this will only work for optimisim

In the case that this if condition returns false, the function returns undefined but it should never
I think a better structure might be 
```
 if (assetInfo.chain === ChainId.Optimism) {
    if(!l1feePrice) { 
      throw new Error('l1FeePrice  must be specified for optimism')
    }
 }
```

This will prevent this function having the `| undefined` return type

monokh (changes requested):

src/store/migrations/21_enable_optimism_chain.ts
skeremidchiev: Not in the Optimism PR! It is already too big and complicated with multiple changes.

monokh:
I think this was continued by another PR

PR #64, opened 2022-06-13
Feat/balance batching

kraikov:
Merge after https://github.com/liquality/wallet-core/pull/63


skeremidchiev (approved):

src/store/actions/updateBalances.ts
```diff
@@ -14,75 +15,90 @@ export const updateBalances = async (
...
-        assets,
-        async (asset) => {
-          let addresses: Address[] = [];
-          const _client = client({
-            network,
-            walletId,
-            asset,
-            accountId: account.id,
-          });
+      const { assets, type, chain } = account;
```
monokh: assets from the account are actually not the assets that was passed in to this action. It would be unexpected for those balances to update

src/utils/arrays.ts
```diff
@@ -0,0 +1,8 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function chunkArray<T = any>(arr: Array<T>, size: number) {
```
monokh: Lodash also has chunk, and we already have that in the lib, should we use that instead?

src/store/actions/updateBalances.ts
monokh: Maybe we should change from asset to chain?

monokh (changes requested):

src/store/actions/updateBalances.ts
```diff
@@ -1,88 +1,110 @@
...
+};
+
+export const updateBalances = async (context: ActionContext, request: UpdateBalanceRequestType) => {
+  const { walletId, network, accountIds } = request;
   const { state, commit, getters } = rootActionContext(context);
+  const getClient = getters.client;
 
-  // filter accounts for all enabled assets
-  let accounts = state.accounts[walletId]?.[network].filter((a) => a.assets && a.assets.length > 0 && a.enabled);
+  const allEnabledAccountIds =
```
monokh: Why calculate all enabled if we're not going to use it (when accoundIds is provided)

monokh (approved):

monokh (approved):

PR #65, opened 2022-06-15
fix: Provide fee asset

AndonMitev:


skeremidchiev (approved):

monokh:
Going to need a description on this one. 
Why is this a requirement now?

AndonMitev:
> Going to need a description on this one. Why is this a requirement now?

Yes so, Astroport dapp works only with UST for fee asset and we were passing Luna soo all dapp interactions were failing, so now wallet is providing the fee asset which is passed from dapp to wallet-core and wallet-core provide correct fee asset to chainify

src/store/actions/sendTransaction.ts
```diff
@@ -30,6 +31,7 @@ export const sendTransaction = async (
     amount: BigNumber;
     data: string;
     fee: number;
+    feeAsset: string;
```
monokh, line 34: Can it be optional so we don't need to specify it when its not required?

monokh (approved):

src/store/actions/sendTransaction.ts
AndonMitev: yes

PR #70, opened 2022-06-23
fix: lower case addresses

skeremidchiev:
Fix for address format issues caused by Rootstock addresses.

### DIFF
--- a/src/swaps/EvmSwapProvider.ts
+++ b/src/swaps/EvmSwapProvider.ts
@@ -38,12 +38,12 @@ export abstract class EvmSwapProvider extends SwapProvider {
     const client = this.getClient(network, walletId, quote.from, quote.fromAccountId) as Client<EvmChainProvider>;
     const signer = client.wallet.getSigner();
 
-    const tokenContract = Typechain.ERC20__factory.connect(fromAsset.contractAddress, signer);
+    const tokenContract = Typechain.ERC20__factory.connect(fromAsset.contractAddress.toLowerCase(), signer);
 
     const userAddressRaw = await this.getSwapAddress(network, walletId, quote.from, quote.fromAccountId);
     const userAddress = chains[fromAsset.chain].formatAddress(userAddressRaw);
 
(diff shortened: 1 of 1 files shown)


src/swaps/EvmSwapProvider.ts
```diff
@@ -38,12 +38,12 @@ export abstract class EvmSwapProvider extends SwapProvider {
     const client = this.getClient(network, walletId, quote.from, quote.fromAccountId) as Client<EvmChainProvider>;
     const signer = client.wallet.getSigner();
 
-    const tokenContract = Typechain.ERC20__factory.connect(fromAsset.contractAddress, signer);
+    const tokenContract = Typechain.ERC20__factory.connect(fromAsset.contractAddress.toLowerCase(), signer);
```
kraikov: No need for that because `fromAsset` comes from the `assetsAdapter` where the lowerCase is performed.

src/swaps/EvmSwapProvider.ts
```diff
@@ -38,12 +38,12 @@ export abstract class EvmSwapProvider extends SwapProvider {
...
 
     const userAddressRaw = await this.getSwapAddress(network, walletId, quote.from, quote.fromAccountId);
     const userAddress = chains[fromAsset.chain].formatAddress(userAddressRaw);
 
-    const allowance = await tokenContract.allowance(userAddress, this.config.routerAddress);
+    const allowance = await tokenContract.allowance(userAddress.toLowerCase(), this.config.routerAddress);
```
kraikov, line 46: no need for that because `formatAddress` from line 44 doesn't perform checksum anymore.

kraikov (changes requested):

src/swaps/EvmSwapProvider.ts
skeremidchiev: Swaps fail if I don't make a lower case on RSK address.
Isn't this checksum happening inside generated code? How exactly it got removed?

kraikov (approved):

PR #72, opened 2022-06-23
Liq 1205 error messaging chainify version

tima-t:
:book LIQ-1205 :book

### DIFF
--- a/src/swaps/SwapProvider.ts
+++ b/src/swaps/SwapProvider.ts
@@ -40,6 +40,11 @@ export abstract class SwapProvider {
    */
   public abstract getSupportedPairs({ network }: { network: Network }): Promise<PairData[]>;
 
+    /**
+   * Get min swap amount for given swap provider
+   */
+  public abstract getMin(quoteRequest: QuoteRequest) : number | any
+
   /**
    * Get a quote for the specified parameters
    */
--- a/src/swaps/SwapProvider.ts
+++ b/src/swaps/SwapProvider.ts
@@ -1,3 +1,4 @@
+import BigNumber from 'bignumber.js';
 import store, { ActionContext } from '../store';
 import { createNotification } from '../store/broker/notification';
 import { MarketData, Network, PairData, SwapHistoryItem } from '../store/types';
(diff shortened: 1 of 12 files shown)


src/swaps/oneinch/OneinchSwapProvider.ts
```diff
@@ -238,6 +238,11 @@ class OneinchSwapProvider extends SwapProvider {
     return null;
   }
 
+  // @ts-ignore
```
monokh: Why the ts-ignore everywhere? going to need a good reason

src/swaps/SwapProvider.ts
```diff
@@ -40,6 +40,11 @@ export abstract class SwapProvider {
    */
   public abstract getSupportedPairs({ network }: { network: Network }): Promise<PairData[]>;
 
+    /**
+   * Get min swap amount for given swap provider
+   */
+  public abstract getMin(quoteRequest: QuoteRequest) : number | any
```
monokh: `| any` 
Why? The possible values for getting the minimum should be number or in the case of no minimum, null or undefined

src/swaps/oneinch/OneinchSwapProvider.ts
kraikov: It's probably to ignore the unused parameter. Better use `_` for the parameter name to ignore the warning, i.e. `_quoteRequest`

tima-t: It is because of the unused argument. 
This function getMin is used in all swap providers and some of them need the argument but others don't. 

Krasi is correct I will use the _ for quoteRequest

src/swaps/SwapProvider.ts
tima-t: I think that liquality provider is returning some kind of object in getMin, but I will check if I can refactor this piece of code to make it consistent

src/swaps/oneinch/OneinchSwapProvider.ts
tima-t: Resolved  ts-ignore removed and _ used

src/swaps/SwapProvider.ts
tima-t: Resolved getMin is returning BigNumber each time, any removed

PR #77, opened 2022-06-28
feat: Sort accounts and assets

AndonMitev:
📚 [Linear Ticket](https://linear.app/liquality/issue/LIQ-1315/ux-for-swaps-the-first-token-appear-is-arbeth-which-is-not-supported) 📚
📚 [Linear Ticket](https://linear.app/liquality/issue/LIQ-481/feedback-can-we-show-assets-in-alphabetical-order-under-chains-on) 📚
📚 [Linear Ticket](https://linear.app/liquality/issue/LIQ-1410/on-disable-native-asset-and-when-user-re-enable-it-gas-token-display#comment-800149fd) 📚


src/store/getters.ts
```diff
@@ -223,29 +225,93 @@ export default {
             enabledChains[activeWalletId]?.[activeNetwork]?.includes(account.chain)
         )
           .map((account) => {
+            /*
+                Calculate fiat balances and asset balances
+                Sort and group assets by dollar value / token amount / market cap
+            */
             const totalFiatBalance = accountFiatBalance(activeWalletId, activeNetwork, account.id);
+            const assetsWithFiat: { asset: string; type: string; amount: BN }[] = [];
```
kraikov: Can we extract `{ asset: string; type: string; amount: BN }` as a separate type?

src/store/actions/updateBalances.ts
```diff
@@ -65,16 +65,16 @@ export const updateBalances = async (context: ActionContext, request: UpdateBala
...
+              _assets.forEach((asset, innerIndex) => {
                 // if balance is `null` there was a problem while fetching
                 const balance = balances[index][innerIndex];
                 if (balance) {
                   commit.UPDATE_BALANCE({
                     network,
                     accountId: account.id,
                     walletId,
-                    asset: asset.code,
+                    asset: assets[innerIndex],
```
kraikov: thanks for fixing this

src/store/getters.ts
```diff
@@ -265,14 +331,26 @@ export default {
     const { state } = rootGetterContext(context);
     const { fiatRates } = state;
     return (asset: AssetType, balance: BigNumber): BigNumber | null => {
-      if (fiatRates && fiatRates[asset] && balance) {
+      if (fiatRates && fiatRates[asset] && balance.gt(0)) {
```
kraikov: shall we check if balance exists? i.e. `balance?.gt(0)`

src/store/getters.ts
```diff
@@ -265,14 +331,26 @@ export default {
...
         return cryptoToFiat(amount, fiatRates[asset]) as BigNumber;
       }
       return null;
     };
   },
+  assetMarketCap(...context: GetterContext) {
+    const { state } = rootGetterContext(context);
+    const { currenciesInfo } = state;
+
+    return (asset: AssetType): BigNumber | null => {
```
kraikov: you can use `Nullable` type from Chainify

kraikov (changes requested):

src/store/actions/updateBalances.ts
AndonMitev: thanks for collaboration

skeremidchiev (approved):

src/store/getters.ts
AndonMitev: done

src/store/getters.ts
AndonMitev: good catch, awesome, thank you! I saw somewhere it return '0' if there is no balance but extra checks doesn't hurt :D

src/store/getters.ts
AndonMitev: applied

src/store/utils.ts
```diff
@@ -105,3 +105,76 @@ export async function getPrices(baseCurrencies: string[], toCurrency: string) {
...
+
+  return coindIds.reduce((acc, currValue) => {
+    const { coinGeckoId, asset } = currValue;
+    const coinInfo = data.find((coin) => coin.id === coinGeckoId);
+
+    return (acc = {
+      ...acc,
+      [asset]: coinInfo ? new BN(coinInfo.market_cap) : new BN(0),
+    });
+  }, {}) as CurrenciesInfo;
```
monokh: As is masking some kind of error here?
Make sure the object created conforms properly to `CurrenciesInfo`

monokh (approved):

src/store/utils.ts
AndonMitev: it's removed

PR #81, opened 2022-06-29
fix: overwrite errors on state transitions in swaps

skeremidchiev:
[LIQ-1278](https://linear.app/liquality/issue/LIQ-1278/swap-is-success-but-display-66-stage-in-red-instead-of-success)
[LIQ-1432](https://linear.app/liquality/issue/LIQ-1432/rsk20-to-avax-swap-fail-with-the-mentioned-error-and-even-after)

### DIFF
--- a/src/store/actions/performNextAction/index.ts
+++ b/src/store/actions/performNextAction/index.ts
@@ -44,6 +44,11 @@ export const performNextAction = async (
     updates = { error: e.toString() };
   }
   if (updates) {
+    if (!updates.error) {
+      // if no error accured overwrite previous errors in history
+      updates.error = null;
+    }
+
     commit.UPDATE_HISTORY({
       network,
       walletId,
@@ -55,6 +60,7 @@ export const performNextAction = async (
       ...item,
       ...updates,
     } as HistoryItem);
+
     if (!updates.error) {
       dispatch.performNextAction({ network, walletId, id });
     }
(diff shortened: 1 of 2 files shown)


src/store/actions/performNextAction/index.ts
```diff
@@ -55,6 +60,7 @@ export const performNextAction = async (
       ...item,
       ...updates,
     } as HistoryItem);
+
```
AndonMitev, line 63: extra line.. :D

src/store/actions/performNextAction/index.ts
```diff
@@ -44,6 +44,11 @@ export const performNextAction = async (
     updates = { error: e.toString() };
   }
   if (updates) {
+    if (!updates.error) {
+      // if no error accured overwrite previous errors in history
+      updates.error = null;
```
AndonMitev, line 49: what is the `error` value? seems like falsy and we are setting it to null?

kraikov (approved):

src/store/actions/performNextAction/index.ts
monokh, line 49: type of error is `error?: string;`
Which I believe means `null` is not valid. Has to be undefined

monokh: Maybe you can move this to the mutation instead.

monokh: ignore that last comment 😄  it might actually cause other issues

kraikov: image

monokh: then should error type also have null as possible value? Because null is not a valid value in `error?: string`

kraikov: https://github.com/liquality/chainify/blob/dev/packages/types/lib/index.ts#L19 this can be used i believe

PR #82, opened 2022-06-30
fix: correctly set the dollar value sign

skeremidchiev:
[LIQ-937](https://linear.app/liquality/issue/LIQ-937/dollar-sign-is-missing-on-dashboard-send-screens-in-case-of-value-is)


kraikov (approved):

kraikov:
image

PR #85, opened 2022-06-30
fix: Updated logic that was causing issues on the mobile app #mob-9

adilmezghouti:
Updated some logic that is causing issues on the mobile app side.

### DIFF
--- a/src/store/mutations.ts
+++ b/src/store/mutations.ts
@@ -146,8 +146,8 @@ export default {
   ) {
     const itemIndex = state.history[network]?.[walletId].findIndex((i) => i.id === id);
     if (itemIndex != undefined && itemIndex >= 0) {
-      const item = state.history[network]?.[walletId][itemIndex]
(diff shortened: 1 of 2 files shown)


src/store/mutations.ts
```diff
@@ -145,9 +145,10 @@ export default {
       updates: Partial<HistoryItem>;
     }
   ) {
-    const item = state.history[network]?.[walletId].find((i) => i.id === id);
-    if (item) {
-      Object.assign(item, updates);
+    const itemIndex = state.history[network]?.[walletId].findIndex((i) => i.id === id);
+    if (itemIndex != undefined && itemIndex >= 0) {
+      const item = state?.history?.[network]?.[walletId]?.[itemIndex]
+      if (item) Vue.set(state.history[network]![walletId], itemIndex, Object.assign(Object.assign({}, item), updates))
```
monokh: isn't that object assign same as doing
`Object.assign({}, item, updates)` ? This seems more readable

monokh (approved):

monokh (approved):

PR #91, opened 2022-07-02
feat: sort chains by dollar value or market cap or alphabetical

AndonMitev:
https://linear.app/liquality/issue/LIQ-1499/asset-display-for-a-brand-new-wallet

### DIFF
--- a/src/store/getters.ts
+++ b/src/store/getters.ts
@@ -22,7 +22,7 @@ import {
   NFTWithAccount,
   WalletId,
 } from './types';
-import { orderAssets } from './utils';
+import { orderAssets, orderChains } from './utils';
 
 const clientCache: { [key: string]: Client } = {};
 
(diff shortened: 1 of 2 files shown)


src/store/getters.ts
```diff
@@ -239,17 +239,22 @@ export default {
...
               const marketCap = assetMarketCap(asset);
               const tokenBalance = account.balances[asset];
-              const { type } = cryptoassets[asset];
+              const { type, matchingAsset } = cryptoassets[asset];
 
               if (fiat) {
                 hasFiat = true;
                 assetsWithFiat.push({ asset, type, amount: fiat });
               } else if (marketCap) {
+                if (type === 'native' && !matchingAsset) {
```
kraikov: you can use the enum here instead of string `native`

kraikov (changes requested):

src/store/getters.ts
AndonMitev: rdy

kraikov (approved):

PR #97, opened 2022-07-25
fix: improve balance fetching for multiaccount

kraikov:


src/store/actions/updateBalances.ts
```diff
@@ -1,108 +1,251 @@
...
+      const result = await client.chain.multicall.multicall<BigNumber[]>(multicallCallData);
+      console.log('result', result);
+
+      accounts.forEach((acc) => {
+        const balances = result.splice(0, acc.assets.length).map((balance, index) => {
+          if (balance) {
+            return balance.toString();
+          } else {
+            console.debug(`Balance not fetched: ${acc.assets[index]}`);
+            return '0';
```
monokh: IMO setting '0' on failure is not the right outcome. It will mean that if there is a network error, user will suddenly see 0 and might even freak out. Can we do something where failed fetch just does not update the balance for that sset ? This way, previous balance will be preserved.

src/store/mutations.ts
```diff
@@ -198,6 +198,50 @@ export default {
...
+
+          const updatedBalances = assets.reduce((result, asset, index) => {
+            result[asset] = balances[index];
+            return result;
+          }, {} as Record<string, string>);
+
+          const updatedAccount = {
+            ...account,
+            balances: { ...currentBalances, ...updatedBalances },
+            loadingInitialBalance: false,
```
monokh: This property is not defined anywhere in the state? We should formally define it or remove it if it's unused.

monokh (changes requested):

src/store/mutations.ts
kraikov: It's а heritage from the `UPDATE_BALANCE` action, that's why I left it

src/store/actions/updateBalances.ts
kraikov: Fixed in https://github.com/liquality/wallet-core/pull/97/commits/f8c7cc7f642eb9b809e4ae0aaf4cd440d8b95d5f

src/store/mutations.ts
kraikov: fixed in https://github.com/liquality/wallet-core/pull/97/commits/f8c7cc7f642eb9b809e4ae0aaf4cd440d8b95d5f

monokh (approved):

PR #98, opened 2022-07-25
fix: typo in swap status was refactored

tima-t:
📚 LIQ-1377 📚


kraikov:
This is an example why we need to use enums.

kraikov (approved):

PR #99, opened 2022-07-25
feat: lifi integration

skeremidchiev:
LiFi integration
[LIQ-1555](https://linear.app/liquality/issue/LIQ-1555/lifi)


src/swaps/lifi/LifiSwapProvider.ts
```diff
@@ -0,0 +1,321 @@
...
+  EstimateFeeRequest,
+  EstimateFeeResponse,
+  NextSwapActionRequest,
+  QuoteRequest,
+  SwapRequest,
+  SwapStatus,
+} from '../types';
+
+export interface LifiSwapProviderConfig extends EvmSwapProviderConfig {
+  rpcURL: string;
```
monokh: Their http service is not like an RPC

src/swaps/lifi/LifiSwapProvider.ts
```diff
@@ -0,0 +1,321 @@
...
+  async getMin(_quoteRequest: QuoteRequest) {
+    return new BN(0);
+  }
+
+  // ======== HELPERS ========
+
+  private getDefaultRouteOptions() {
+    return {
+      slippage: 5 / 100, // 5%
+      order: Orders[0], // 'RECOMMENDED'
```
monokh: This loooks super weird but ok

src/swaps/lifi/LifiSwapProvider.ts
```diff
@@ -0,0 +1,321 @@
...
+      const bestRoute = await this._lifiClient.getQuote(quoteRequest);
+      return { from, to, fromAmount: fromAmountInUnit, toAmount: bestRoute.estimate.toAmount, bestRoute: bestRoute };
+    } catch (e) {
+      console.warn(e);
+      return null;
+    }
+  }
+
+  async newSwap(swap: SwapRequest<LifiSwapHistoryItem>) {
+    const approvalAddress = swap.quote.bestRoute?.estimate.approvalAddress;
```
monokh: `bestRoute?` might become an issue.
In this case, if a bestRoute does not exist, this is serious error that shuld be thrown rather than silently setting `approvalAddress = undefined`

`if(!swap.quote.bestRoute) throw new Error('....')`
before is better. Then it doesn't need`?.`

src/swaps/lifi/LifiSwapProvider.ts
```diff
@@ -0,0 +1,321 @@
...
+      bridge: quote.bestRoute?.tool,
+      fromChain: this.getChainByID(quote.bestRoute?.action.fromChainId as number),
+      toChain: this.getChainByID(quote.bestRoute?.action.toChainId as number),
+      txHash: quote.fromFundHash,
+    });
+
+    return result;
+  }
+
+  private getChainByID(id: number) {
```
monokh: Can you explain what this does? Not very clear. There's a lot of casting between where it's called and inside this function and it's making me worried because the types are not being enforced at all

src/swaps/lifi/LifiSwapProvider.ts
```diff
@@ -0,0 +1,321 @@
...
+    const indexOfFromChainId = Object.values(ChainId).indexOf(id as unknown as ChainId);
+    return Object.keys(ChainId)[indexOfFromChainId];
+  }
+
+  // lifi type of swaps are simillar to boost (combination of multiple single and/or cross chain swaps)
```
monokh: In this case, are all lifi swaps only 1 wallet transaction? worth checking...

src/swaps/liqualityboost/liqualityBoostERC20toNative/LiqualityBoostERC20toNative.ts
```diff
@@ -238,7 +238,7 @@ class LiqualityBoostERC20toNative extends SwapProvider {
       return {
         ...resultLSP,
         ...swapLSP,
-        toAmount: resultLSP.toAmount,
+        toAmount: (resultLSP as any).toAmount,
```
monokh: why all the `as any` all of a sudden 😅

monokh (changes requested):

package.json
```diff
@@ -32,6 +32,7 @@
     "@ethersproject/address": "5.6.0",
     "@ethersproject/providers": "5.6.0",
     "@hop-protocol/sdk": "0.0.1-beta.314",
+    "@lifi/sdk": "^1.0.2",
```
kraikov: can we lock the version?

src/store/actions/getQuotes.ts
```diff
@@ -17,7 +17,16 @@ export const getQuotes = async (
     toAccountId,
     // Amount is string because in some contexts, it is passed over messages not supporting full objects
     amount,
-  }: { network: Network; from: Asset; to: Asset; fromAccountId: AccountId; toAccountId: AccountId; amount: string }
+    walletId,
+  }: {
```
kraikov: can we create a type for this? it's hard to read?

src/swaps/EvmSwapProvider.ts
```diff
@@ -43,29 +43,58 @@ export abstract class EvmSwapProvider extends SwapProvider {
     const userAddressRaw = await this.getSwapAddress(network, walletId, quote.from, quote.fromAccountId);
     const userAddress = chains[fromAsset.chain].formatAddress(userAddressRaw);
 
-    const allowance = await tokenContract.allowance(userAddress.toLowerCase(), this.config.routerAddress);
-
-    const isLSP = swapRequest.quote.provider === SwapProviderType.Liquality;
+    const _routerAddress = approvalAddress ?? this.config.routerAddress;
```
kraikov: why _ ?

src/swaps/EvmSwapProvider.ts
```diff
@@ -43,29 +43,58 @@ export abstract class EvmSwapProvider extends SwapProvider {
...
+      const { quote, network, walletId } = swapRequest;
+      const fromAsset = assetsAdapter(quote.from)[0];
+
+      // only ERC20 tokens allowed
+      const client = this.getClient(network, walletId, quote.from, quote.fromAccountId) as Client<EvmChainProvider>;
+      const signer = client.wallet.getSigner();
+
+      const tokenContract = Typechain.ERC20__factory.connect(String(fromAsset.contractAddress), signer);
+      const approveAmount = approveMax ? ethers.constants.MaxUint256 : ethers.BigNumber.from(quote.fromAmount);
+      const _routerAddress = approvalAddress ?? this.config.routerAddress;
```
kraikov: why _ ?

src/swaps/lifi/LifiSwapProvider.ts
```diff
@@ -0,0 +1,321 @@
...
+  rpcURL: string;
+}
+
+export interface LifiSwapHistoryItem extends EvmSwapHistoryItem {
+  fromFundHash: string;
+  fromFundTx: Transaction<EvmTypes.EthersTransactionResponse>;
+}
+
+class LifiSwapProvider extends EvmSwapProvider {
+  readonly config: LifiSwapProviderConfig;
```
kraikov: good usage of readonly, can we also specify the visibility modifier? same for the rest of the class variables

src/swaps/lifi/LifiSwapProvider.ts
```diff
@@ -0,0 +1,321 @@
...
+  readonly _lifiClient: LiFi;
+  private _httpClient: HttpClient;
+  readonly nativeAssetAddress: string = '0x0000000000000000000000000000000000000000';
```
kraikov: we can use constants from ethers for the 0th address

src/swaps/lifi/LifiSwapProvider.ts
```diff
@@ -0,0 +1,321 @@
...
+    } catch (e) {
+      console.warn(e);
+      return null;
+    }
+  }
+
+  async newSwap(swap: SwapRequest<LifiSwapHistoryItem>) {
+    const approvalAddress = swap.quote.bestRoute?.estimate.approvalAddress;
+    const approveTx = await this.approve(swap, false, approvalAddress);
+    const updates = approveTx || (await this.initiateSwap(swap));
```
kraikov: even if the syntax is good, it's hard to read and it takes a while to get what's going on...

can we have the more ugly, but more readable version:

```ts
if(approveTx) {
...
} else {
...
}
```

src/swaps/lifi/LifiSwapProvider.ts
```diff
@@ -0,0 +1,321 @@
...
+    return { id: uuidv4(), ...updates };
+  }
+
+  private async initiateSwap({ network, walletId, quote }: SwapRequest<LifiSwapHistoryItem>) {
+    const client = super.getClient(network, walletId, quote.from, quote.fromAccountId);
+    const txData: TransactionRequest = quote.bestRoute?.transactionRequest as TransactionRequest;
+    const fromFundTx = await client.wallet.sendTransaction(txData);
+
+    return {
+      status: 'WAITING_FOR_INITIATION_CONFIRMATIONS',
```
kraikov, line 140: PLS, let's use the PR to introduce enums for the statuses

src/swaps/lifi/LifiSwapProvider.ts
kraikov: to expand on that, since the `swap.quote.bestRoute` is used on multiple places, maybe a getter would fit better...
e.g.

```ts
getBestRoute(swapQuote) {
  if(!quote.bestRoute) {
    throw...
  } else {
    ...
  }
}
```

src/swaps/lifi/LifiSwapProvider.ts
```diff
@@ -0,0 +1,321 @@
...
+
+  async getMin(_quoteRequest: QuoteRequest) {
+    return new BN(0);
+  }
+
+  // ======== HELPERS ========
+
+  private getDefaultRouteOptions() {
+    return {
+      slippage: 5 / 100, // 5%
```
kraikov: can we directly use the result of `5/100` instead of dividing every time?

kraikov (changes requested):

src/swaps/lifi/LifiSwapProvider.ts
skeremidchiev: `getChainByID` method takes `ChainID` and returns the key of following enum: https://github.com/lifinance/types/blob/3ba8bb7712adcd87a2023f655cd6c1f2c48b253e/src/base.ts

src/swaps/lifi/LifiSwapProvider.ts
skeremidchiev: What do you mean by 1 wallet? Especially when we talk about Liquality wallet?
There are 3 types of swaps:
1) 'Swap' -> single chain (like uniswap)
2) 'Cross' -> multi chain (bridging like hop for example)
3) 'LiFi' -> some combination of the previous 2 types (you can have 2 swaps or 3 swaps and 1 cross )

src/swaps/liqualityboost/liqualityBoostERC20toNative/LiqualityBoostERC20toNative.ts
skeremidchiev: Have to change it to `LiqualitySwapHistoryItem`.

src/swaps/lifi/LifiSwapProvider.ts
skeremidchiev: This will be resolved in other PR.

src/swaps/lifi/LifiSwapProvider.ts
monokh: I mean if they are 1 transaction. I  am wondering if some of the lifi swaps might require more than 1 transaction from the user and if our provider is supporting that correctly.

monokh (approved):

kraikov (approved):
Once the conflict is resolved, LGTM

PR #100, opened 2022-07-27
feat: Solana-jupiter-integration

AndonMitev:


package.json
```diff
@@ -35,6 +35,8 @@
     "@hop-protocol/sdk": "0.0.1-beta.314",
     "@liquality/cryptoassets": "1.13.4",
     "@metamask/eth-sig-util": "^4.0.1",
+    "@solana/spl-token": "^0.2.0",
+    "@solana/web3.js": "^1.48.0",
```
kraikov: can we lock the versions?

src/swaps/jupiter/JupiterSwapProvider.ts
```diff
@@ -0,0 +1,243 @@
...
+      info: {
+        ...info,
+        toAddress,
+      },
+    };
+  }
+
+  public async newSwap(quoteInput: SwapRequest<JupiterSwapHistoryItem>) {
+    const connection = new Connection(
+      'https://solana--mainnet.datahub.figment.io/apikey/d7d9844ccf72ad4fef9bc5caaa957a50',
```
kraikov: use https://github.com/liquality/wallet-core/blob/master/src/utils/networks.ts#L121 pls

kraikov: you can also use

`const connection = client.chain.getProvider()` to reuse the connection made in chainify

src/store/actions/accounts/createAccount.test.ts
```diff
@@ -13,7 +13,7 @@ test('accounts createAccount', async () => {
     key: '0x1234567890123456789012345678901234567890',
   });
   const walletId = wallet.state.activeWalletId;
-  expect(wallet.state.accounts[walletId!]?.mainnet.length).toBe(11);
+  expect(wallet.state.accounts[walletId!]?.mainnet.length).toBe(13);
```
monokh: You are adding one account but the test checks 2 more accounts?

src/swaps/jupiter/JupiterSwapProvider.ts
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  // Jupiter on their site before performing swap are recommending:
+  // "We recommend having at least 0.05 SOL for any transaction"
+  // https://jup.ag/swap/SOL-USDT
+
+  // Here we are returning `33333333` Lamports because in wallet we multiply by 1.5 which
+  // output exactly 0.05 SOL
+  getExtraAmountToExtractFromBalance() {
+    return 33333333;
```
kraikov, line 193: this doesn't look good... what's the idea here?

src/store/migrations/21_enable_solana_chain.ts
```diff
@@ -0,0 +1,69 @@
...
+  version: 21,
+  migrate: async (state: any) => {
+    const accounts: any = {};
+    const enabledChains: any = {};
+    for (const walletId in state.accounts) {
+      accounts[walletId] = {};
+      enabledChains[walletId] = {};
+
+      for (const network of Networks) {
+        const accountExistsAndProperlyInitialized = state.accounts[walletId][network].find(
```
monokh: Why would this ever be the case? The migration only runs from a version that solana was not added.

kraikov (changes requested):

src/store/migrations/21_enable_solana_chain.ts
```diff
@@ -0,0 +1,69 @@
...
+              index: 0,
+              derivationPath,
+              color: getNextAccountColor(ChainId.Solana, 0),
+            },
+          });
+          accounts[walletId][network] = [...state.accounts[walletId][network], solanaAccount];
+        }
+
+        const chainEnabled = state.enabledChains[walletId][network].includes(ChainId.Solana);
+        if (chainEnabled) {
```
monokh: Again, this case shouldn't exist

src/store/migrations/21_enable_solana_chain.ts
```diff
@@ -0,0 +1,69 @@
...
+        }
+      }
+    }
+
+    const enabledAssets: any = {};
+    for (const network of Networks) {
+      enabledAssets[network] = {};
+      for (const walletId in state.enabledAssets[network]) {
+        enabledAssets[network][walletId] = [...state.enabledAssets[network][walletId]];
+        if (!enabledAssets[network][walletId].includes('SOL')) enabledAssets[network][walletId].push('SOL');
```
monokh: as above

src/swaps/jupiter/JupiterSwapProvider.ts
```diff
@@ -0,0 +1,243 @@
...
+    const { from, to, amount } = quoteRequest;
+    const fromInfo = cryptoassets[from];
+    const toInfo = cryptoassets[to];
+
+    // only for Solana network swaps
+    if (fromInfo.chain !== ChainId.Solana || toInfo.chain !== ChainId.Solana || amount.lt(0)) {
+      return null;
+    }
+
+    const solMintAddress = 'So11111111111111111111111111111111111111112';
```
monokh: this should be constant at the top?

src/swaps/jupiter/JupiterSwapProvider.ts
```diff
@@ -0,0 +1,243 @@
...
+
+    /* Full swap flow might be constructed from 3 different transactions:
+      1. Setup transaction  - used to handle creating ATA or Serum open order accounts
+      2. Swap transaction  - performing the actual swaps
+      3. Clean up transaction - unwrap SOL if a SOL swap.
+      You must send each transaction sequentially in the order:
+      setupTransaction -> swapTransaction -> cleanupTransaction and wait for each to be 'confirmed' before sending the next one.
+    */
+
+    for (let i = 0; i < txTypes.length; i++) {
```
monokh, line 136: Because the transactions are really fast, they don't need to be in different steps?

monokh (changes requested):
Any erc20 tokens on testnet to test for solana?

monokh (commented):
Also any jupiter testnet?

src/swaps/jupiter/JupiterSwapProvider.ts
AndonMitev: described in the comment above the function

> Also any jupiter testnet?

no :(

AndonMitev: image
 They recommend to have this amount for possible fees available, comment above the function, i believe this is because swap might consist 1-3 transactions and they don't have a way to tell exactly how many transactions will be thats why they want from user to have higher available balance for possibly fees when using Max

AndonMitev:
> Any erc20 tokens on testnet to test for solana?

jupiter operates only on mainnet

AndonMitev:
> Also any jupiter testnet?

no, only mainnet

src/swaps/jupiter/JupiterSwapProvider.ts
AndonMitev: They are in different steps:

`You must send each transaction sequentially in the order:
      setupTransaction -> swapTransaction -> cleanupTransaction and wait for each to be 'confirmed' before sending the next one.`

src/swaps/jupiter/JupiterSwapProvider.ts
kraikov: 33333333 * 1.5 = 49999999,5 😆 

where is the multiplication of 1.5? can we skip it and set 0.05 here instead?

src/store/migrations/21_enable_solana_chain.ts
AndonMitev: I have just copy and paste Avalanche migration and adjust for Solana :D

src/store/migrations/21_enable_solana_chain.ts
AndonMitev: I have just copy and paste Avalanche migration and adjust for Solana :D

src/store/migrations/21_enable_solana_chain.ts
AndonMitev: I have just copy and paste Avalanche migration and adjust for Solana :D

package.json
AndonMitev: Resolved 🚀

src/swaps/jupiter/JupiterSwapProvider.ts
AndonMitev: Resolved 🚀

src/store/actions/accounts/createAccount.test.ts
AndonMitev: Resolved 🚀

src/store/migrations/21_enable_solana_chain.ts
AndonMitev: Resolved 🚀

src/store/migrations/21_enable_solana_chain.ts
AndonMitev: Resolved 🚀

src/store/migrations/21_enable_solana_chain.ts
AndonMitev: Resolved 🚀

src/swaps/jupiter/JupiterSwapProvider.ts
AndonMitev: Resolved 🚀

src/swaps/jupiter/JupiterSwapProvider.ts
AndonMitev: Yes we add to this fee which is static and it is - 00000000.5 SOL,

Here is the multiplyer: 

https://github.com/liquality/wallet/blob/2ca93e41dfc52efe5f55659e1b14956849f17b6e/src/views/Swap/Swap.vue#L884

AndonMitev: Maybe i can pass some boolean to the wallet and skip multiplication for Solana chain not sure if this is good tho

kraikov: i think its better to have some leftover then

this is only when sending max, right?

AndonMitev: This left over is kept every time, let's say you have 2 SOL, Max available will be 1.95

AndonMitev: If you want to send just 1 SOL, no problem you send 1, just reducing the max amount and keep some for extra fee if needed

AndonMitev: I think there are 2 reasons for this:

1. Jupiter is not aware how many txs will be broadcasted 1-3
2. Tx for creation of token accounts, which are part of this jupiter transactions

monokh (approved):

PR #101, opened 2022-07-27
fix: add minimum for BTC thorchain swaps

monokh:
Fixes LIQ-1558

### DIFF
--- a/src/swaps/thorchain/ThorchainSwapProvider.ts
+++ b/src/swaps/thorchain/ThorchainSwapProvider.ts
@@ -47,6 +47,10 @@ const OUT_MEMO_TO_STATUS = {
   REFUND: 'REFUNDED',
 };
 
+const MINIMUM_SWAP_AMOUNTS: { [asset: string]: string } = {
+  BTC: '0.00010001',
+};
+
 /**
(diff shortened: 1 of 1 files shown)


src/swaps/thorchain/ThorchainSwapProvider.ts
```diff
@@ -504,7 +504,10 @@ class ThorchainSwapProvider extends SwapProvider {
     return null;
   }
 
-  async getMin(_quoteRequest: QuoteRequest) {
+  async getMin(quote: QuoteRequest) {
+    if (quote.from === 'BTC') { // Thorchain swaps have to be more than 10000 sats
+      return new BN(0.00010001)
```
kraikov: Do you believe it's a good idea to have a mapping that says:
`[BTC]: BN(0.00010001)`

so we can easily extend later if needed?


the usage would be

`return map[quote.from] || BN(0)`

monokh: sure

bradleySuira (approved):
LGTM

kraikov (approved):

PR #105, opened 2022-07-29
fix: Thorchain min and slippage to prevent refunds

monokh:


src/swaps/thorchain/ThorchainSwapProvider.ts
```diff
@@ -267,14 +259,36 @@ class ThorchainSwapProvider extends SwapProvider {
...
+
+  async getQuote(quoteRequest: QuoteRequest) {
+    const { from, to, amount } = quoteRequest;
+    // Only ethereum, bitcoin and bc chains are supported
+    if (!SUPPORTED_CHAINS.includes(cryptoassets[from].chain) || !SUPPORTED_CHAINS.includes(cryptoassets[to].chain))
+      return null;
+
+    // Slippage must be higher for small value swaps
+    const min = await this.getMin(quoteRequest);
+    const slippage = new BN(amount).gt(min.times(2)) ? 0.03 : 0.05;
```
kraikov, line 279: why min * 2?
why 3 & 5%?

monokh: `min * 2` is just a qualifier for a small vs bigger swap size
The default slippage is 3% (I found this suggestion from asgardex), 5% is a slippage to give smaller value swaps a higher chance of succeeding
I feel the comment above it describes this in short

kraikov (approved):

PR #107, opened 2022-08-04
feat: adding a getSendAmountFee function

johannafransn:
Fixes part of MOB-55 by adding a getSendAmountFee() function to wallet core, so we can use this on both wallet-extension and wallet-mobile
- fix


src/utils/coinFormatter.ts
```diff
@@ -52,6 +52,5 @@ export const formatFiat = (amount: BN) => {
 };
 
 export const formatFiatUI = (amount: number) => {
-  const _amount = String(amount).replace(/,/g, '');
```
monokh: I'm worried that wallet might still be passing a string here even though the type we have defined is number :( 
can we keep this part?

monokh (approved):
Just a minor issue

monokh:
Check also the PR title check fail

monokh (approved):

PR #109, opened 2022-08-05
feat: fast btc withdraw (rbtc->btc) provider

monokh:
Implementing RBTC -> BTC using fastbtc bridge
Fixes LIQ-1578

### DIFF
--- a/src/swaps/fastbtc/FastBTCWithdrawSwapProvider.ts
+++ b/src/swaps/fastbtc/FastBTCWithdrawSwapProvider.ts
@@ -1,0 +1,394 @@
+import { BitcoinBaseWalletProvider, BitcoinEsploraApiProvider, BitcoinTypes } from '@chainify/bitcoin';
+import { Client } from '@chainify/client';
+import { EvmChainProvider, EvmTypes, EvmWalletProvider } from '@chainify/evm';
+import { Transaction } from '@chainify/types';
+import { JsonRpcProvider } from '@ethersproject/providers';
+import { ChainId, chains, currencyToUnit, unitToCurrency } from '@liquality/cryptoassets';
+import BN from 'bignumber.js';
+import * as ethers from 'ethers';
+import { v4 as uuidv4 } from 'uuid';
+import { buildConfig } from '../..';
+import { ActionContext } from '../../store';
+import { withInterval } from '../../store/actions/performNextAction/utils';
+import { Network, SwapHistoryItem, WalletId } from '../../store/types';
+import { prettyBalance } from '../../utils/coinFormatter';
+import cryptoassets from '../../utils/cryptoassets';
+import { ChainNetworks } from '../../utils/networks';
+import { SwapProvider } from '../SwapProvider';
+import {
+  BaseSwapProviderConfig,
+  EstimateFeeRequest,
+  EstimateFeeResponse,
+  NextSwapActionRequest,
+  QuoteRequest,
+  SwapQuote,
+  SwapRequest,
+  SwapStatus,
+} from '../types';
+
+export interface FastBtcWithdrawSwapHistoryItem extends SwapHistoryItem {
+  transferId: string;
+  swapTxHash: string;
+  swapTx: Transaction<EvmTypes.EthersTransactionResponse>;
+  receiveTxHash: string;
+  receiveTx: Transaction<BitcoinTypes.Transaction>;
+}
+
+export interface FastBtcWithdrawSwapProviderConfig extends BaseSwapProviderConfig {
+  network: Network;
+  routerAddress: string;
+}
+
+const BitcoinTransfer =
+  '(address rskAddress, uint8 status, uint8 nonce, uint8 feeStructureIndex, uint32 blockNumber, uint40 totalAmountSatoshi, string btcAddress)';
+
+const ABI = [
+  'event NewBitcoinTransfer(bytes32 indexed transferId, string btcAddress, uint256 nonce, uint256 amountSatoshi, uint256 feeSatoshi, address indexed rskAddress)',
+  'event BitcoinTransferStatusUpdated(bytes32 indexed transferId, uint8 newStatus)',
+  'event BitcoinTransferBatchSending(bytes32 bitcoinTxHash, uint8 transferBatchSize)',
+  'function minTransferSatoshi() view returns (uint40)',
+  'function maxTransferSatoshi() view returns (uint40)',
+  'function calculateCurrentFeeSatoshi(uint256 amountSatoshi) public view returns (uint256)',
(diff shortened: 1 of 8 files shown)


src/swaps/fastbtc/FastBTCWithdrawSwapProvider.ts
```diff
@@ -0,0 +1,394 @@
...
+
+export interface FastBtcWithdrawSwapProviderConfig extends BaseSwapProviderConfig {
+  network: Network;
+  routerAddress: string;
+}
+
+const BitcoinTransfer =
+  '(address rskAddress, uint8 status, uint8 nonce, uint8 feeStructureIndex, uint32 blockNumber, uint40 totalAmountSatoshi, string btcAddress)';
+
+const ABI = [
```
adilmezghouti: [Pebble] Can we pull this to a separate file or read it from the blockchain?

src/swaps/fastbtc/FastBTCWithdrawSwapProvider.ts
```diff
@@ -0,0 +1,394 @@
...
+  'event BitcoinTransferBatchSending(bytes32 bitcoinTxHash, uint8 transferBatchSize)',
+  'function minTransferSatoshi() view returns (uint40)',
+  'function maxTransferSatoshi() view returns (uint40)',
+  'function calculateCurrentFeeSatoshi(uint256 amountSatoshi) public view returns (uint256)',
+  'function transferToBtc (string calldata btcAddress) external payable',
+  `function getTransferByTransferId(bytes32 transferId) public view returns (${BitcoinTransfer})`,
+];
+
+/// @dev Status of an rBTC-to-BTC transfer.
+enum BitcoinTransferStatus {
```
adilmezghouti: [Pebble] It would be nice to export these types as well in case we need them on the front end.

src/swaps/fastbtc/FastBTCWithdrawSwapProvider.ts
```diff
@@ -0,0 +1,394 @@
...
+  nonce: number; // each Bitcoin address can be reused up to 255 times
+  feeStructureIndex: number; // the fee calculation to be applied to this transfer
+  blockNumber: number; // the RSK block number where this was initialized
+  totalAmountSatoshi: number; // the number of BTC satoshis that the user sent
+  btcAddress: string; // the BTC address in legacy or Bech32 encoded format
+}
+
+class FastBTCWithdrawSwapProvider extends SwapProvider {
+  config: FastBtcWithdrawSwapProviderConfig;
+  _provider: JsonRpcProvider;
```
adilmezghouti: [Pebble] if you mean to convey that _provider is private by using the underscore then we need to declare it private as well. We should also look into making it a coding standard within our teams.

src/swaps/fastbtc/FastBTCWithdrawSwapProvider.ts
```diff
@@ -0,0 +1,394 @@
...
+    const fastBtcBridge = this.getFastBtcBridge(this._provider);
+    const fromAmountInSatoshi = new BN(currencyToUnit(cryptoassets.BTC, new BN(amount)));
+    const feeSatoshi: ethers.BigNumber = await fastBtcBridge.calculateCurrentFeeSatoshi(
+      ethers.BigNumber.from(fromAmountInSatoshi.toFixed())
+    );
+    const fromAmountInUnit = currencyToUnit(cryptoassets[from], new BN(amount));
+    const toAmountInUnit = fromAmountInSatoshi.minus(new BN(feeSatoshi.toString()));
+
+    return {
+      // TODO: Why is from amount being set in every quote call? It should be in one place only (getQuotes action)
```
adilmezghouti, line 114: [Pebble] should we create a ticket for this todo item so we don't lose track of it?

src/swaps/fastbtc/FastBTCWithdrawSwapProvider.ts
```diff
@@ -0,0 +1,394 @@
...
+      fromAmount: fromAmountInUnit.toFixed(),
+      toAmount: toAmountInUnit.toFixed(),
+    };
+  }
+
+  async buildSwapTx({ network, walletId, quote }: { network: Network; walletId: WalletId; quote: SwapQuote }) {
```
adilmezghouti: [Pebble] I think it would be better to avoid inline type declaration for ease of reuse.

src/swaps/fastbtc/FastBTCWithdrawSwapProvider.ts
```diff
@@ -0,0 +1,394 @@
...
+      const gasPrice = new BN(feePrice).times(1e9); // ETH fee price is in gwei
+      const fee = new BN(gasLimit.toString()).times(1.1).times(gasPrice);
+      fees[feePrice] = unitToCurrency(cryptoassets.RBTC, fee);
+    }
+
+    return fees;
+  }
+
+  /**
+   * TODO: this is called very often from clients, so making a network call each time will be very intensive
```
adilmezghouti, line 186: [Pebble] We need to create a ticket for it for better tracking

adilmezghouti (approved):

src/swaps/fastbtc/FastBTCWithdrawSwapProvider.ts
monokh: Can't read this from blockchain, solidity code is not stored on there - its metadata. 
moved to separate file

src/swaps/fastbtc/FastBTCWithdrawSwapProvider.ts
monokh: Yes exported and added to separate file

src/swaps/fastbtc/FastBTCWithdrawSwapProvider.ts
monokh: made private

src/swaps/fastbtc/FastBTCWithdrawSwapProvider.ts
monokh: We should but practically I know we will never pick it up and it will just be more headache for the PM

src/swaps/fastbtc/FastBTCWithdrawSwapProvider.ts
monokh: fixed

src/swaps/fastbtc/FastBTCWithdrawSwapProvider.ts
monokh: We should but practically I know we will never pick it up and it will just be more headache for the PM

PR #110, opened 2022-08-08
feat: unstoppable domain integration 

amritkumarj:


package.json
```diff
@@ -42,6 +42,7 @@
     "@uniswap/v2-core": "^1.0.1",
     "@uniswap/v2-periphery": "^1.1.0-beta.0",
     "@uniswap/v2-sdk": "^3.0.1",
+    "@unstoppabledomains/resolution": "^8.3.1",
```
kraikov: can you please pin the version

kraikov (changes requested):

src/utils/uns.ts
```diff
@@ -0,0 +1,96 @@
...
+                return 'RSK'
+            case ChainId.Ethereum:
+            case ChainId.Arbitrum:
+            default:
+                return 'ETH'
+        }
+
+    }
+
+    async lookupDomain(address: string, chainId: ChainId): Promise<string | null> {
```
kraikov: Can you use `Nullable<string>` instead of `string | null`?

src/utils/uns.ts
```diff
@@ -0,0 +1,97 @@
+import { HttpClient } from '@chainify/client';
```
monokh: `utils` seems the wrong place for this
how about `src/nameResolvers/uns.ts` ?

src/build.config.ts
```diff
@@ -208,6 +211,9 @@ const config: WalletCoreConfig = {
     ChainId.Avalanche,
   ],
   supportedBridgeAssets: ['MATIC', 'RBTC', 'AVAX'],
+  resolutionService : 'https://unstoppabledomains.g.alchemy.com/domains/',
```
monokh: This should be grouped in some way
perhaps?
```
nameResolvers: {
  uns: {
    resolutionService: ...,
    tldAPI: ...,
    alchemyKey: ...,
  }
}
```

monokh (changes requested):
Looking good!
just some small comment

monokh (approved):

PR #119, opened 2022-08-12
fix: fastbtc withdraw, no quote when not in range

monokh:
Fixes LIQ-1578

### DIFF
--- a/packages/wallet-core/src/swaps/fastbtc/FastBTCWithdrawSwapProvider.ts
+++ b/packages/wallet-core/src/swaps/fastbtc/FastBTCWithdrawSwapProvider.ts
@@ -102,6 +102,10 @@ class FastBTCWithdrawSwapProvider extends SwapProvider {
       return null;
     }
 
+    const { min, max } = await this.getLimits();
+    const isQuoteAmountInTheRange = amount.gte(min) && amount.lte(max);
+    if (!isQuoteAmountInTheRange) return null;
+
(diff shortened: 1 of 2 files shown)


packages/wallet-core/src/swaps/fastbtc/FastBTCWithdrawSwapProvider.ts
```diff
@@ -102,6 +102,10 @@ class FastBTCWithdrawSwapProvider extends SwapProvider {
       return null;
     }
 
+    const { min, max } = await this.getLimits();
+    const isQuoteAmountInTheRange = amount.gte(min) && amount.lte(max);
+    if (!isQuoteAmountInTheRange) return null;
```
kraikov: can we use brackets for one line statements, please?

kraikov (changes requested):

AndonMitev (approved):

AndonMitev (approved):

packages/wallet-core/src/swaps/fastbtc/FastBTCWithdrawSwapProvider.ts
AndonMitev: Yes pleasee, devs like me do not read after closing bracket on if statements  if ( condition ) <- Thats the end line :D

monokh: Sure, wonder if there is a prettier/lint config for this?

PR #121, opened 2022-08-17
refactor: fee calculations

skeremidchiev:
Refactor Fees
[LIQ-1617](https://linear.app/liquality/issue/LIQ-1617/fee-estimation-of-nft-is-incorrect)
[LIQ-1273](https://linear.app/liquality/issue/LIQ-1273/max-send-eth-not-working)


src/utils/fees.ts
```diff
@@ -46,46 +60,132 @@ function isEIP1559Fees(chain: ChainId) {
...
+  return suggestedGasFee.suggestedBaseFeePerGas || 0 + suggestedGasFee.maxPriorityFeePerGas;
+}
 
-async function getSendAmountFee(accountId: AccountId, asset: Asset, amount?: BN) {
-  const getMax = amount === undefined
-  const sendFees: {
-    [speed in keyof FeeDetails]: BN
-  } = { slow: new BN(0), average: new BN(0), fast: new BN(0), }
+/*
+ * Max fee for EIP1559
```
johannafransn: Nice, thanks for breaking everything up into smaller functions that you can easily call and display the values :)

johannafransn (approved):
Looks good to me!

src/store/getters.ts
```diff
@@ -201,11 +201,11 @@ export default {
       return account;
     };
   },
-  assetFees(...context: GetterContext) {
+  getSuggestedFeePrices(...context: GetterContext) {
```
monokh: The getters are not usually named this way?
We should be consistent

src/utils/fees.ts
```diff
@@ -46,46 +60,132 @@ function isEIP1559Fees(chain: ChainId) {
   return chain === ChainId.Ethereum || chain === ChainId.Polygon || chain === ChainId.Avalanche;
 }
 
+/*
+ * Probable fee for EIP1559 (BASE + TIP)
+ */
+function probableFeePerUnitEIP1559(suggestedGasFee: EIP1559Fee) {
+  return suggestedGasFee.suggestedBaseFeePerGas || 0 + suggestedGasFee.maxPriorityFeePerGas;
```
monokh: `0 + ` is a quirky way of I presume converting a string to a number? Can we make it more explicit such as `Number.parseFloat()`

src/utils/fees.ts
```diff
@@ -46,46 +60,132 @@ function isEIP1559Fees(chain: ChainId) {
...
+function feePerUnit(suggestedGasFee: FeeType, chain: ChainId): number {
+  if (suggestedGasFee === undefined || chain === undefined) {
+    throw new Error('feePerUnit: Incorrect input!');
+  }
+
+  /*
+   * In case it is EIP1559 chain then `maxFeePerGas` will be set.
+   * Otherwise its non EIP1559 (this includes both EVM and Non EVM chains) and calculations are done based on `fee`.
+   */
+  if (isEIP1559Fees(chain) && (<EIP1559Fee>suggestedGasFee).maxFeePerGas) {
```
monokh: There is way too much casting in this function

This is FeeType:
`export type FeeType = EIP1559Fee | number;`

Meaning you can first check if `suggestedGasFee` is a number and return,
Otherwise it is implicitly type `EIP1559Fee` and typescript will recognise it as such as long as other condition (number) has been handled before it

I believe this would remove all the casting in this function

src/utils/fees.ts
```diff
@@ -46,46 +60,132 @@ function isEIP1559Fees(chain: ChainId) {
...
+    _sendFees[_speed] = _sendFees[_speed].plus(getSendFee(feeAsset, _fee));
+  }
+
+  return _sendFees;
+}
+
+/*
+ * Send fee estimation method for BTC
+ */
+async function sendTxFeesInBTC(
```
monokh: The name of this function is focusing on the unit returned, `InBTC` This gives the impression that there are other function returning values in different units 
IMO better name would be `sendBitcoinTxFees`

monokh (changes requested):

src/utils/fees.ts
skeremidchiev: Not really. `suggestedBaseFeePerGas` is optional property for `EIP1559Fee`. In case `suggestedBaseFeePerGas` is undefined (which highly likely), then it is replaced by 0.

src/utils/fees.ts
skeremidchiev: Ok, will remove casting.

src/utils/fees.ts
skeremidchiev: `sendTxFeesInBTC` -> `sendBitcoinTxFees`
But then I have to rename: `sendTxFeesInNativeAsset` to have consistent names for all methods?
Any proposal?

src/utils/fees.ts
kraikov: logical operators have greater priority than arithmetic operators, not sure if the prettier won't get rid of ( ) if you use them for better readability.

src/store/getters.ts
skeremidchiev: Removing leading `get` out of the method's name.

src/utils/fees.ts
monokh: > In case suggestedBaseFeePerGas is undefined (which highly likely), then it is replaced by 0.

Ok that's even more unpredictable than I expected then 😅 
Is there a way to write this so I can understand you are expecting 0 when it's undefined

skeremidchiev: Done

skeremidchiev: Done

kraikov:
In the future it'll be good if the `BigNumber` is used from a single place (e.g. it's exported from Chainify).
If wallet-core is using a different version from Chainify at some point there could be an incompatibility issues.
It's not part of this PR, but must be considered for the future.
cc @monokh @johannafransn

kraikov (approved):

monokh (approved):

AndonMitev (approved):

PR #138, opened 2022-08-30
feat: refactor migrations that enable chains

skeremidchiev:
This PR refactors migrations that enable chains in order to reduce code duplication.

### DIFF
--- a/packages/wallet-core/src/store/migrations/enable_chain.ts
+++ b/packages/wallet-core/src/store/migrations/enable_chain.ts
@@ -5,8 +5,6 @@ import { Networks } from '../../utils/networks';
 import { AccountType } from '../types';
 
 async function enableChain(state: any, chainId: ChainId) {
-  console.log('enableChain: ', chainId);
-
   const accounts: any = {};
   const enabledChains: any = {};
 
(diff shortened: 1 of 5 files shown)


packages/wallet-core/src/store/migrations/enable_chain.ts
```diff
@@ -0,0 +1,76 @@
...
+      if (chainEnabled) {
+        enabledChains[walletId][network] = [...state.enabledChains[walletId][network]];
+      } else {
+        enabledChains[walletId][network] = [...state.enabledChains[walletId][network], chainId];
+      }
+    }
+  }
+
+  const enabledAssets: any = {};
+  for (const network of Networks) {
```
kraikov, line 52: is it possible to use the above networks loop?

kraikov: you can have 2 methods -> `handleAccounts` & `handleAssets` both of them can accept network and you can do it in a single loop

packages/wallet-core/src/store/migrations/enable_chain.ts
```diff
@@ -0,0 +1,76 @@
+import { ChainId, getChain } from '@liquality/cryptoassets';
+import { accountCreator, getNextAccountColor } from '../../utils/accounts';
+import { getDerivationPath } from '../../utils/derivationPath';
+import { Networks } from '../../utils/networks';
+import { AccountType } from '../types';
+
+async function enableChain(state: any, chainId: ChainId) {
+  console.log('enableChain: ', chainId);
```
kraikov: Remove.

kraikov (changes requested):

kraikov (approved):

monokh (approved):

PR #166, opened 2022-09-08
feat: NFT tx notification

Jennievon:
- update NFTs after send txn.
- create history notification for NFT txn.


packages/wallet-core/src/store/broker/notification.ts
```diff
@@ -52,7 +73,17 @@ const createSendNotification = (item: SendHistoryItem) => {
   });
 };
 
+const createSendNFTNotification = (item: NFTSendHistoryItem) => {
+  if (!(item.status in NFT_SEND_STATUS_MAP)) return;
```
kraikov: pls, use { } for one liners

packages/wallet-core/src/store/broker/notification.ts
```diff
@@ -52,7 +73,17 @@ const createSendNotification = (item: SendHistoryItem) => {
...
+
+  return createNotification({
+    ...notification,
+  });
+};
+
 export const createHistoryNotification = (item: HistoryItem) => {
   if (item.type === 'SEND') return createSendNotification(item);
   else if (item.type === 'SWAP') return createSwapNotification(item);
+  else if (item.type === 'NFT') return createSendNFTNotification(item);
```
kraikov: pls, use { } for one liners

kraikov (changes requested):

packages/wallet-core/src/store/mutations.ts
```diff
@@ -131,6 +131,15 @@ export default {
 
     state.history[network]![walletId].push(transaction);
   },
+  REMOVE_NFT_FROM_STATE(
```
monokh: This is not correctly named
All mutations do something to the state, so `FROM_STATE` should not be required

packages/wallet-core/src/store/actions/sendNFTTransaction.ts
```diff
@@ -13,8 +13,11 @@ export const sendNFTTransaction = async (
   const account: Account = getters.accountItem(accountId)!;
   const asset = getChain(network, account.chain).nativeAsset[0].code;
   const client = getters.client({ network, walletId, chainId: account.chain });
+
   const tx = await client.nft.transfer(nft.asset_contract!.address!, receiver, [nft.token_id!], values);
 
+  commit.REMOVE_NFT_FROM_STATE({ network, walletId, accountId, nft });
```
monokh: This method is innaccurate because it does not portray the real status of NFTs in your wallet. 
Let's assume you make the `send` call, and the tx does not confirm or gets dropped. Your NFT will still be gone from state. 

Instead, the accurate way to do this is to refresh the list of NFTs after a `sendNFT` tx has been confirmed
This should be in `sendNFT` `performNextAction`

monokh (changes requested):

monokh:
Please add a changeset

.changeset/stale-candles-rush.md
```diff
@@ -0,0 +1,5 @@
+---
+'@liquality/wallet-core': major
```
monokh: A patch or minor changeset should be enough for this. No breaking changes are being made

packages/wallet-core/src/store/actions/performNextAction/index.ts
```diff
@@ -39,6 +39,7 @@ export const performNextAction = async (
         walletId,
         transaction: item,
       });
+      dispatch.updateNFTs({ network, walletId });
```
monokh: It seems wasteful here to update the entire wallet NFTs
It should update only the single account that was used to send

monokh: Also why is this here? this will get called everytime an update happens, but you only need to update the state after the NFT send tx is confirmed

packages/wallet-core/src/store/actions/updateNFTs.ts
```diff
@@ -36,6 +37,9 @@ export const updateNFTs = async (
       const nfts = nftAssetsFetched.map((nftAsset: NFT) => {
         const nftAssetStoredInState = nftAssetsStoredInState.find((asset: NFT) => asset.token_id === nftAsset.token_id);
         const starred = nftAssetStoredInState ? nftAssetStoredInState.starred : false;
+        // const image = nftAsset.image_original_url.includes('ipfs://')
```
monokh: comment?

packages/wallet-core/src/store/actions/updateNFTs.ts
```diff
@@ -1,20 +1,21 @@
...
 export const updateNFTs = async (
   context: ActionContext,
   {
     walletId,
     network,
-    accountIds,
   }: {
     walletId: WalletId;
     network: Network;
-    accountIds: AccountId[];
```
monokh: Why is this changing? It should still be possible to update only a single account's NFTs

monokh (changes requested):

packages/wallet-core/src/store/actions/updateNFTs.ts
```diff
@@ -36,6 +36,19 @@ export const updateNFTs = async (
       const nfts = nftAssetsFetched.map((nftAsset: NFT) => {
         const nftAssetStoredInState = nftAssetsStoredInState.find((asset: NFT) => asset.token_id === nftAsset.token_id);
         const starred = nftAssetStoredInState ? nftAssetStoredInState.starred : false;
+        const replaceIPFSUrl = (url: string) => {
```
monokh: Wasn't this supposed to go in chainify? 
https://github.com/liquality/chainify/pull/612

packages/wallet-core/src/store/actions/performNextAction/sendNFT.ts
```diff
@@ -19,13 +19,17 @@ async function waitForConfirmations(
...
 ): Promise<Partial<NFTSendHistoryItem> | undefined> {
-  const { getters } = rootActionContext(context);
+  const { getters, dispatch } = rootActionContext(context);
   const { from, accountId } = transaction;
   const chainId = getAsset(network, from).chain;
   const client = getters.client({ network, walletId, chainId, accountId });
   try {
     const tx = await client.chain.getTransactionByHash(transaction.txHash);
     if (tx && tx.confirmations && tx.confirmations > 0) {
+      if (transaction.type === 'NFT') {
```
monokh: Can the type be anything else? It's inside `sendNFT`

packages/wallet-core/src/store/actions/performNextAction/sendNFT.ts
```diff
@@ -19,13 +19,17 @@ async function waitForConfirmations(
...
+        dispatch.updateNFTs({ network, walletId, accountIds: [transaction.accountId] });
```
monokh: needs to await so that process does not finish before updating

monokh (changes requested):

packages/wallet-core/src/store/actions/updateNFTs.ts
Jennievon: I have moved it there

monokh (approved):

PR #187, opened 2022-09-20
fix: price source should be omitted when matching currency code is there

kraikov:


monokh (approved):

monokh:
some TS error
also needs changeset? 
Otherwise LGTM

kraikov:
> some TS error also needs changeset? Otherwise LGTM

Can we create a single release once https://github.com/liquality/wallet-core/pull/188 also gets merged?

PR #197, opened 2022-09-26
feat: debridge swap provider

igumirov:
This PR includes:

deBridge swap provider integration in wallet-core package

### DIFF
--- a/packages/wallet-core/src/swaps/debridge/DeBridgeSwapProvider.ts
+++ b/packages/wallet-core/src/swaps/debridge/DeBridgeSwapProvider.ts
@@ -19,15 +19,13 @@ import {
   SwapStatus,
 } from '../types';
 import cryptoassets from '../../utils/cryptoassets';
-import { Asset, Network, SwapHistoryItem, SwapProviderType, WalletId } from '../../store/types';
+import { Asset, Network, SwapHistoryItem, WalletId } from '../../store/types';
 import { isChainEvmCompatible, isERC20 } from '../../utils/asset';
 import { ChainNetworks } from '../../utils/networks';
 import { Transaction } from '@chainify/types';
 import { EvmChainProvider, EvmTypes } from '@chainify/evm';
 import { ActionContext } from '../../store';
-import { Client, HttpClient } from '@chainify/client';
-import { getSwapProvider } from '../../factory';
-import { OneinchSwapProvider } from '../oneinch/OneinchSwapProvider';
+import { Client } from '@chainify/client';
 
 interface BuildSwapQuote extends SwapQuote {
   fee?: number;
@@ -37,6 +35,15 @@ interface BuildSwapRequest {
   network: Network;
   walletId: WalletId;
   quote: BuildSwapQuote;
+}
+
+interface FullSubmissionInfo {
+  claim: {
+    transactionHash: string;
+  };
+  send: {
+    isExecuted: boolean;
+  };
 }
 
 export interface DebridgeSwapHistoryItem extends SwapHistoryItem {
@@ -58,8 +65,8 @@ export interface DebridgeSwapProviderConfig extends BaseSwapProviderConfig {
 }
 
 const zeroAddress = '0x0000000000000000000000000000000000000000';
-const nativeAddress = '0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee';
 const slippagePercentage = 1;
+const slippageBps = slippagePercentage * 100;
 const chainIds = [1, 56, 137, 42161, 43114];
 
 class DeBridgeSwapProvider extends SwapProvider {
@@ -96,7 +103,7 @@ class DeBridgeSwapProvider extends SwapProvider {
         params: {
           srcChainId: chainIdFrom,
           srcChainTokenIn: fromToken,
-          srcChainTokenInAmount: fromAmountInUnit.toNumber(),
+          srcChainTokenInAmount: fromAmountInUnit.toFixed(),
           slippage: slippagePercentage,
           dstChainId: chainIdTo,
           dstChainTokenOut: toToken,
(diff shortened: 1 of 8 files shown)


packages/wallet-core/src/swaps/debridge/DeBridgeSwapProvider.ts
```diff
@@ -0,0 +1,616 @@
...
+    try {
+      const fromToken = cryptoassets[from].contractAddress || zeroAddress;
+      const toToken = cryptoassets[to].contractAddress || zeroAddress;
+      const trade = await axios({
+        url: this.config.url + 'estimation',
+        method: 'get',
+        params: {
+          srcChainId: chainIdFrom,
+          srcChainTokenIn: fromToken,
+          srcChainTokenInAmount: fromAmountInUnit.toNumber(),
```
monokh: numbers in JS have a max that could be too small. Better to use `toFixed()` to get string here

packages/wallet-core/src/swaps/debridge/DeBridgeSwapProvider.ts
```diff
@@ -0,0 +1,616 @@
...
+          from: approvalTx.from,
+          to: approvalTx.to,
+          data: approvalTx.data,
+          value: '0x' + approvalTx.value!.toString(16),
+        };
+
+        gasLimit += (await client.chain.getProvider().estimateGas(rawApprovalTx)).toNumber();
+      }
+
+      const swapTx: any = await this.loadSwapTx({ network, walletId, quote });
```
monokh: why any? Looks like this type from the function is well defined

packages/wallet-core/src/swaps/debridge/DeBridgeSwapProvider.ts
```diff
@@ -0,0 +1,616 @@
...
+        from: fromAddress,
+        to: swapTx.to,
+        value: '0x' + Number(swapTx.value).toString(16),
+        data: swapTx.data,
+      };
+
+      try {
+        gasLimit += (await client.chain.getProvider().estimateGas(rawSwapTx)).toNumber();
+      } catch (e) {
+        if (swapTx.preSwap) {
```
monokh: could you add a little comment explaining what is happening here? 
If gas estimation fails, one inch is used?

packages/wallet-core/src/swaps/debridge/DeBridgeSwapProvider.ts
```diff
@@ -0,0 +1,616 @@
...
+
+  protected _toTxType(): string | null {
+    return null;
+  }
+
+  protected _totalSteps(): number {
+    return 4;
+  }
+
+  protected _timelineDiagramSteps(): string[] {
```
monokh, line 543: on the wallet, the timeline is strange. This is a MATIC -> BNB swap:
image

It should say Collecting BNB or Swapping MATIC
It's probably a bug in the wallet

packages/wallet-core/src/swaps/debridge/DeBridgeSwapProvider.ts
```diff
@@ -0,0 +1,616 @@
...
+        const chainIdFrom = ChainNetworks[cryptoassets[quote.from].chain][network].chainId;
+        const debridgeGate = new ethers.Contract(
+          this.config.chains[chainIdFrom as number].deBridgeGateAddress,
+          DeBridgeGate.abi,
+          provider
+        );
+        globalFee = (await debridgeGate.globalFixedNativeFee()) || 0;
+      }
+
+      const i = new BigNumber(21e3);
```
monokh: Could you label the number and variable? `21e3` I assume this number is a send gas limit?

packages/wallet-core/src/swaps/debridge/DeBridgeSwapProvider.ts
monokh: If it's possible, it would be good to also add the receive transaction. 
Hop and Thorchain swap providers demonstrate how this can be done. Then the timeline will display the swap tx as well as the transaction where the output was received.

packages/wallet-core/src/swaps/debridge/DeBridgeSwapProvider.ts
```diff
@@ -0,0 +1,616 @@
...
+  public async newSwap({ network, walletId, quote }: BuildSwapRequest) {
+    const approvalRequired = isERC20(quote.from);
+    const updates = approvalRequired
+      ? await this.approveTokens({ network, walletId, quote })
+      : await this.sendSwap({ network, walletId, quote });
+
+    return {
+      id: uuidv4(),
+      fee: quote.fee,
+      slippage: 50,
```
monokh: shouldn't this be according to `slippagePercentage` ? 
1 percent would be 100 (bips)

packages/wallet-core/src/swaps/debridge/DeBridgeSwapProvider.ts
```diff
@@ -0,0 +1,616 @@
...
+        url: this.config.url + 'transaction',
+        method: 'get',
+        params: {
+          srcChainId: chainIdFrom,
+          srcChainTokenIn: fromToken,
+          srcChainTokenInAmount: quote.fromAmount,
+          slippage: slippagePercentage,
+          dstChainId: chainIdTo,
+          dstChainTokenOut: toToken,
+          dstChainTokenOutRecipient: fromAddress,
```
monokh: Token out account could differ from the token in account. 
When users select assets to swap, they also explicitly select an account, this could be a different account from the default. 
To get the correct address: 
`await this.getSwapAddress(network, walletId, quote.from, quote.toAccountId);`

packages/wallet-core/src/swaps/debridge/DeBridgeSwapProvider.ts
```diff
@@ -0,0 +1,616 @@
...
+          dstChainFallbackAddress: fromAddress,
```
monokh: it would be same as above I think

packages/wallet-core/src/swaps/debridge/DeBridgeSwapProvider.ts
```diff
@@ -0,0 +1,616 @@
...
+        },
+      });
+      const estimation = result?.data?.estimation;
+      return result?.data?.tx
+        ? {
```
monokh: The few lines below is hard to read 😅 
Is it possible to split it across a few statements, so it's easier to read and follow?

packages/wallet-core/src/swaps/debridge/DeBridgeSwapProvider.ts
```diff
@@ -0,0 +1,616 @@
...
+    }
+  }
+
+  private async waitForSwapExecution({ swap, network, walletId }: NextSwapActionRequest<DebridgeSwapHistoryItem>) {
+    try {
+      if (await this.isExecuted(swap.swapTxHash)) {
+        const client = this.getClient(network, walletId, swap.from, swap.fromAccountId);
+        const { status } = await client.chain.getProvider().getTransactionReceipt(swap.swapTxHash);
+        this.updateBalances(network, walletId, [swap.fromAccountId]);
+        return {
```
monokh, line 478: Here is where you would set the receiveTx if we could do that

monokh (changes requested):
Thanks for contribution ✨ 
I tested and it worked. Just some small issues.

monokh (approved):

PR #199, opened 2022-09-28
feat: add chainify error parser and other changes

koderholic:
This PR does the following

1. Adds an Error Parser for Chainify
2. Updates Wallet Core so that a number of calls to chainify are wrapped using the Chainify Error Parser
3. Updates Wallet Core so that calls to oneInch APIs are wrapped using OneInch Error Parsers
4. Updates Wallet Core so that quote calls to Lifi are wrapped using LifiErrorParser
5. Minor updates to Liquality Error structure
6. Captures Custom Errors(thrown in wallet-core and wallet extension) in error-parser and refactor wallet-core so that Custom Errors are replaced with Liquality Errors.
7. Updates to error reporting

Link to Linear:
https://linear.app/liquality/issue/EXT-36/error-and-exception-handling-error-service-qa-engineer


packages/error-parser/src/reporters/console.ts
```diff
@@ -1,4 +1,4 @@
-import { LiqualityError } from '../LiqualityErrors';
+import { LiqualityError } from '../LiqualityErrors/LiqualityError';
 
 export function reportToConsole(error: LiqualityError) {
   console.log(error);
```
kraikov: Remove

kraikov (approved):

packages/error-parser/src/reporters/console.ts
koderholic: Hey @kraikov why do you suggest to remove?

kraikov: Sorry, I saw a console.log and my immediate reaction was to be removed.

koderholic: Oh got it!

packages/wallet-core/src/store/actions/sendTransaction.ts
```diff
@@ -43,15 +44,20 @@ export const sendTransaction = async (
...
-    to: getChain(network, chainId).formatAddress(to),
-    value: new BN(amount),
-    data,
-    gasLimit: gas,
-    fee,
-    asset: _asset,
-    feeAsset: _feeAsset,
-  });
+  const parser = getErrorParser(ChainifyErrorParser);
+  const tx = (await parser.wrapAsync(
```
monokh: Is there any way to make the wrapping of chainify calls seamless? 
It strikes me as verbose and error prone to have to wrap every call individually. 
Instead you could return proxies that handle chainify errors when client is created in `clients.ts`

packages/error-parser/src/parsers/wrapCustomError.ts
```diff
@@ -0,0 +1,8 @@
+import { InternalError } from '../LiqualityErrors';
+
+export function wrapCustomError(customError: any) {
```
monokh: this makes more sense as `createCustomError` 
when you read it in usage
`throw wrapCustomError(..)`
vs
`throw createCustomError(...)`

monokh: It would be even better if this function didn't exist and instead `InternalError` took an optional `rawError` in constructor. 

Then usage is even more simple:
`throw new InternalError({ raw: CUSTOM_ERRORS.NotFound.LedgerTransportCreator })`

monokh: by proxy i mean js proxy: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy

packages/wallet-core/src/store/actions/sendTransaction.ts
koderholic: Done

packages/error-parser/src/parsers/wrapCustomError.ts
koderholic: Done

packages/wallet-core/src/utils/getTransactionByHash.ts
```diff
@@ -0,0 +1,6 @@
+import { Client } from '@chainify/client';
+import { Transaction } from '@chainify/types';
+
+export async function getTransactionByHash(client: Client, hash: string): Promise<Transaction<any>> {
```
monokh: What's the purpose of this extra level of indirection?
Just makes it a bit harder to figure out what the actual call is

Could we remove function and just call directly?

monokh (approved):

packages/wallet-core/src/utils/getTransactionByHash.ts
koderholic: Oh! was handy in the initial approach of wrapping each call to chainify to reduce verbosity resulting from the wraps. I'll just remove this now. Thanks for the mention.

koderholic: Done

PR #211, opened 2022-10-12
fix: assets must be fetched from store instead of cryptoassets

kraikov:


packages/wallet-core/src/store/actions/updateFees.ts
```diff
@@ -8,7 +7,7 @@ export const updateFees = async (context: ActionContext, { asset }: { asset: Ass
   const network = state.activeNetwork;
   const walletId = state.activeWalletId;
 
-  const chainId = getAsset(network, asset).chain;
+  const chainId = getters.cryptoassets[asset].chain;
```
bradleySuira, line 10: I just think that maybe we can handle the case when we don't have getters.cryptoassets[asset]

bradleySuira (approved):
LGTM

packages/wallet-core/src/store/actions/updateFees.ts
kraikov: I agree. However there shouldn't be a case in which an asset is used and it's missing in the store. If that happens, there is a bigger, greater issue that must be fixed.

monokh (approved):

monokh (approved):

PR #223, opened 2022-10-19
fix: `analyticsEnabled`  return value

Jennievon:


packages/wallet-core/src/store/getters.ts
```diff
@@ -378,7 +378,7 @@ export default {
   },
   analyticsEnabled(...context: GetterContext): boolean {
     const { state } = rootGetterContext(context);
-    if (state.analytics && state.analytics.acceptedDate != null) {
+    if (state.analytics && state.analytics.acceptedDate != null && state.analytics.acceptedDate > 0) {
       return true;
     }
     return false;
```
mpwanyi256: @Jennievon We could re factor lines 381 to 384 to `return !!(state.analytics && state.analytics.acceptedDate != null && state.analytics.acceptedDate > 0)`

It would always default to true or false

Jennievon: sure! this works

mpwanyi256 (approved):
@Jennievon CODE LGTM 🔥

koderholic (approved):

PR #230, opened 2022-10-25
fix: Solana address checker bug

koderholic:


packages/cryptoassets/src/chains/NonEvmChain.ts
```diff
@@ -15,7 +16,16 @@ export abstract class NonEvmChain extends EvmChain {
 
 export class SolanaChain extends NonEvmChain {
   public isValidAddress(address: string) {
-    return typeof address === 'string' && address.length >= 32 && address.length <= 44;
+    try {
+      const isValidLength = address.length >= 32 && address.length <= 44;
+      if (!address || !isValidLength) {
+        return false;
+      }
+      new PublicKey(address); // throws if address is invalid
```
mpwanyi256: @koderholic Great job, however, just for my own clarity, I was under the impression that to validate the address, we would use the the PublicKey constructor along with the `isOnCurve` method. Something like this;

```
const owner = new PublicKey("DS2tt4BX7YwCw7yrDNwbAdnYrxjeCPeGJbHmZEYC8RTb");
console.log(PublicKey.isOnCurve(owner.toBytes())); // true
console.log(PublicKey.isOnCurve(owner.toString())); // true
```

koderholic: @mpwanyi256 I thought that too. Stack overflow and personal experiment revealed that not all Solana addresses pass the isOnCurve check. Solana addresses(accounts) are of two types: Executable (Akin to smart contract address) and non-Executable.  Only non-executable addresses pass the isOnCurve check. However all addresses pass the conversion to public key using the PublicKey. So adding the isOnCurve check will ultimately limit our function to `isValidNonExecutableSolanaAddress` instead of more generally `isValidSolanaAddress`

mpwanyi256: @koderholic Makes sense. Thanks

mpwanyi256 (approved):
@koderholic CODE LGTM 🔥

koderholic:
@mpwanyi256 After you looked through, I made a tiny update to further simplify the function.

packages/cryptoassets/src/chains/NonEvmChain.ts
```diff
@@ -15,7 +15,16 @@ export abstract class NonEvmChain extends EvmChain {
...
 export class SolanaChain extends NonEvmChain {
   public isValidAddress(address: string) {
-    return typeof address === 'string' && address.length >= 32 && address.length <= 44;
+    try {
+      const PUBLIC_KEY_LENGTH = 32;
+
+      const publicKey = base58.decode(address);
+      if (publicKey.length !== PUBLIC_KEY_LENGTH) return false;
+
+      return true;
```
mpwanyi256: @koderholic We can just return `(publicKey.length === PUBLIC_KEY_LENGTH)`

koderholic: True! Thanks @mpwanyi256

PR #240, opened 2022-11-07
fix: resolve issue with transaction status not updating for send and swap

koderholic:
fix: issue with transaction status not updating properly
fix: add a source field to the error report object, to identify the source of the error

Reference ticket here:
https://linear.app/liquality/issue/COR-76/high-all-eth-transactions-are-displayed-as-failed-in-wallet
https://linear.app/liquality/issue/LIQ-1709/swap-failed-for-optimism-chain-using-lifi-rates-on-stage1-and-on-retry
https://linear.app/liquality/issue/COR-77/common-error-message-displaying-in-the-wallet-regardless-of-the-issue

### DIFF
--- a/.changeset/calm-eagles-suffer.md
+++ b/.changeset/calm-eagles-suffer.md
@@ -1,7 +1,0 @@
----
-'@liquality/cryptoassets': major
-'@liquality/error-parser': major
-'@liquality/wallet-core': major
----
-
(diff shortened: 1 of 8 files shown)


mpwanyi256 (commented):
@koderholic CODE LGTM :fire:

kraikov (approved):
LGTM

PR #244, opened 2022-11-10
Solana injection wallet standart

kraikov:


packages/solana-wallet-injection/src/wallet.ts
```diff
@@ -0,0 +1,184 @@
...
+    );
+  }
+
+  public async connect(
+    _options?: { onlyIfTrusted?: boolean | undefined } | undefined
+  ): Promise<{ publicKey: PublicKey }> {
+    const { accepted } = await this.enable();
+
+    if (!accepted) {
+      throw new Error('User rejected');
```
koderholic, line 51: @kraikov Is it possible to use the error parser here?. [See example](https://github.com/liquality/wallet/blob/cd1982dc5bc9d38f0ed1e7e8e89f9667d4f58755/src/pageProvider/solanaProvider.js#:~:text=if%20(!accepted)%20throw%20new%20UserDeclinedError())

packages/solana-wallet-injection/src/wallet.ts
```diff
@@ -0,0 +1,184 @@
...
+  public async disconnect(): Promise<void> {
+    this.publicKey = null;
+    this.isConnected = false;
+  }
+
+  public async signAndSendTransaction<T extends Transaction | VersionedTransaction>(
+    transaction: T,
+    options?: SendOptions | undefined
+  ): Promise<{ signature: string }> {
+    console.debug('signAndSendTransaction');
```
koderholic, line 74: @kraikov Is there an intermittent bug you're still trying to track down by these console statements?

koderholic (commented):
@kraikov Here are two(2) comments from me atm.

packages/solana-wallet-injection/src/wallet.ts
```diff
@@ -0,0 +1,184 @@
...
+    console.debug('signAndSendTransaction: signedRawTx', signedRawTx);
+    const signature = await this.connection.sendRawTransaction(signedRawTx, options);
+    console.debug('signAndSendTransaction: signature', signature);
+    return { signature };
+  }
+
+  public async signTransaction<T extends Transaction | VersionedTransaction>(transaction: T): Promise<any> {
+    console.debug('signTransaction');
+    const instructions = this.getTransactionDetails(transaction);
+    const secretKeyBase58 = await this.callMethod(COMMON_REQUEST_MAP.wallet_getSigner, {
```
monokh, line 85: This is bringing secrets into the browser's page context, isn't this dangerous? Allowing applications that are using the injection to potentially sniff it out? 

Normally, all required data for the transaction is passed into the wallet and signing happens in the background of the wallet. This ensures that private keys stay sandboxed in the wallet.

packages/solana-wallet-injection/src/wallet.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  // TODO: to be tested
+  public async signMessage(message: Uint8Array): Promise<{ signature: Uint8Array }> {
+    console.debug('signMessage', message);
+    console.debug('signMessage: string', message.toString());
+    return this.callMethod(COMMON_REQUEST_MAP.wallet_signMessage, message.toString());
+  }
+
+  public on<E extends keyof LiqualityEvent>(_event: E, _listener: LiqualityEvent[E], _context?: any): void {
+    console.debug('on', _event);
```
monokh, line 124: So there are event listeners part of the spec. Are we implementing it? If not, what is going to be incompatible?

packages/solana-wallet-standard/src/register.ts
```diff
@@ -0,0 +1,69 @@
+// This is copied from @wallet-standard/wallet
```
monokh, line 1: Is it not possible to import these from that package?

packages/solana-wallet-injection/src/wallet.ts
monokh: Another issue that this brings up is what the user is prompted on the wallet. 
In this approach, they will agree to export their private key?

Injection requests should prompt the specific transaction

monokh (changes requested):

packages/solana-wallet-standard/src/register.ts
kraikov: don't know, followed the guide here: https://github.com/solana-labs/wallet-standard/blob/master/WALLET.md

packages/solana-wallet-injection/src/wallet.ts
kraikov: Without this approach the impact will be huge:
- chainify must introcuce a new interface that will potentially impact other chains
- the solana package (in chainify) must be rewritten from scratch to use the latest solana web3js version

packages/solana-wallet-injection/src/wallet.ts
kraikov: usually injection is very problematic from my experience with the rest of the injection implemnetations

I left these so they can be easily debugged by QA

packages/solana-wallet-injection/src/wallet.ts
kraikov: I don't know. 
I'm testing here as stated in their official guide:
https://solana-labs.github.io/wallet-adapter/example/

it seems to work

packages/solana-wallet-injection/src/wallet.ts
monokh: In this case I don't think we can avoid it - key security is essential

> chainify must introcuce a new interface that will potentially impact other chains

This has been done before, see bitcoin's `signPSBT`, you can follow it through: 
https://github.com/liquality/wallet/blob/cd1982dc5bc9d38f0ed1e7e8e89f9667d4f58755/src/store/modules/app/requestPermission.js#L102
It's fairly separate and shouldn't affect other things. Just an extra interface on solana provider

packages/solana-wallet-standard/src/register.ts
monokh: Wow ok, they got a very rough approach for this at the moment it seems

monokh: didn't realise it was a clone

packages/solana-wallet-injection/src/wallet.ts
monokh: Ok, if it's not part of the test harness, then I guess we wait and see if it gets utilised.

PR #246, opened 2022-11-16
fix: replace default errors with specific errors

koderholic:
What this PR does

1. Added a ledger error parser so that ledger device errors showing up as internal errors would now be more specific
2. Added a node error parser so that RSK's low gas fee error is correctly interpreted. (More node error cases will be added to the parser as they are discovered)
3. Added error parser for Debridge. wrap debridge quote and transaction api calls using the parser
4. Added  basic error parser for thorchain. Wrapped thorchain api calls using the parser

Reference ticket here:
https://linear.app/liquality/issue/COR-77/common-error-message-displaying-in-the-wallet-regardless-of-the-issue
https://linear.app/liquality/issue/LIQ-1632/getting-error-when-swapping-rbtc-to-sov-using-sovryn-rates-for-small
https://linear.app/liquality/issue/COR-75/implement-debridge-error-parser
https://linear.app/liquality/issue/COR-41/implement-thorchain-api-error-parser


monokh (approved):
really need to move messages out of wallet-core though

PR #257, opened 2022-11-30
fix: report error with user consent

koderholic:
This PR does the following

1. Add error reporting config to error parser so that errors are only reported if useReporter is true (based on user preference in wallet)
2. Add action, mutation and migration script to support logging error to state


packages/wallet-core/src/store/migrations/23_add_error_log.ts
```diff
@@ -0,0 +1,11 @@
+import { updateErrorReporterConfig, LiqualityErrorJSON } from '@liquality/error-parser';
+import store from '..';
+
+// Add error log to state
+export const addErrorLog = {
+  version: 23,
+  migrate: async (state: any) => {
+    updateErrorReporterConfig({ fallback: (error: LiqualityErrorJSON) => store.dispatch.logError(error) });
+    return { ...state, errorLog: [] };
```
monokh: I don't think you need a migration for this? 
It's better to avoid migrations unless they are really necessary
This can be achieved by just creating the array if it does not exist in the mutation

packages/wallet-core/src/store/types.ts
```diff
@@ -202,6 +203,7 @@ export type HistoryItem = NFTSendHistoryItem | SendHistoryItem | SwapHistoryItem
 
 export enum ExperimentType {
   ManageAccounts = 'manageAccounts',
+  ReportErrors = 'reportErrors',
```
monokh, line 206: Why does this need to be an experiement? Seems pretty straightforward

monokh: experiements are for when you need to enable something in the experiements tab in the wallet

packages/wallet-core/src/store/types.ts
```diff
@@ -256,6 +258,8 @@ export interface RootState {
   experiments: Partial<Record<ExperimentType, boolean>>;
   whatsNewModalVersion: string;
   enabledChains: WalletIdNetworkMap<ChainId[]>;
+
+  errorLog: LiqualityErrorJSON[];
```
monokh, line 262: Why here? This is the account object in the state. I don't see any errors being specifically stored for the account - only see it at root

packages/wallet-core/src/store/types.ts
monokh: But this feature is necessary and we want to it work at all times
If a user misses an error in the log, makes everyone's life more difficult

monokh (changes requested):

packages/wallet-core/src/store/types.ts
koderholic: Yeah, got your point. See Rydmac's reason for wanting the experiment (https://discord.com/channels/745324936607957022/1047117622816280607/1047213446082613339)

One way to satisfy that desire is to add a variable to env for environment so that when environment is dev, errors would always be reported (discord, sentry, console, e.t.c) What do you think ?

Also, the current implementation logs error to state only when analytics is disabled, do you think it should always log to state ?

packages/wallet-core/src/store/types.ts
koderholic: @monokh Not sure I understand your point. Currently there is one errorLog and it's in the state's root. Do you mean each account should have its own errorLog?

It's possible the collapsing of lines might be causing you to see `errorLog: LiqualityErrorJSON[];` in account object in state rather than in the state root. Could you look again in the full file (https://github.com/liquality/wallet-core/blob/9e5ff10c78e238f1d929f797dba906d5043f8f0f/packages/wallet-core/src/store/types.ts) ?

monokh (approved):

PR #283, opened 2023-01-20
fix: thorchain hash

mpwanyi256:
This PR fixes the thorchain hash address, removing `ox` from transaction hashes if they exist

### DIFF
--- a/packages/wallet-core/src/swaps/thorchain/ThorchainSwapProvider.ts
+++ b/packages/wallet-core/src/swaps/thorchain/ThorchainSwapProvider.ts
@@ -188,8 +188,10 @@ class ThorchainSwapProvider extends SwapProvider {
 
   async _getTransaction(hash: string): Promise<ThorchainTransactionResponse | null> {
     try {
-      return await this.thorchainAPIErrorParser.wrapAsync(() => this._httpClient.nodeGet(`/thorchain/tx/${hash}`), {
-        txHash: hash,
+      const hasOx = hash.charAt(0) + hash.charAt(1) === '0x'
+      const transactionHash = hasOx ? hash.slice(2) : hash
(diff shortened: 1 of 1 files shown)


packages/wallet-core/src/swaps/thorchain/ThorchainSwapProvider.ts
```diff
@@ -188,8 +188,10 @@ class ThorchainSwapProvider extends SwapProvider {
 
   async _getTransaction(hash: string): Promise<ThorchainTransactionResponse | null> {
     try {
-      return await this.thorchainAPIErrorParser.wrapAsync(() => this._httpClient.nodeGet(`/thorchain/tx/${hash}`), {
-        txHash: hash,
+      const hasOx = hash.charAt(0) + hash.charAt(1) === '0x'
+      const transactionHash = hasOx ? hash.slice(2) : hash
+      return await this.thorchainAPIErrorParser.wrapAsync(() => this._httpClient.nodeGet(`/thorchain/tx/${transactionHash}`), {
+        txHash: transactionHash,
```
koderholic, line 194: @mpwanyi256 Is there a reason you're taking out the '0x'?

koderholic, line 194: @mpwanyi256 Is there a reason you're taking out the '0x'?

koderholic (changes requested):

packages/wallet-core/src/swaps/thorchain/ThorchainSwapProvider.ts
mpwanyi256: @koderholic yes, apparently thorchain updated to remove 0x that's why we the api was failing

koderholic: Oh! Got it!

koderholic (approved):
LGTM

